PR #113, opened 2022-01-25
Configure base path

otbe:
## Motivation
Merging this PR enables the eventcatalog user to specify the basePath of the deployment.
Right now we're able to build a static export and deploy it to `example.com` but its not easy to deploy it to `exmaple.com/my-catalog`.

After the merge you can add `basePath` to your `eventcatalog.config.js` and next will pick it up properly.

I think (!) its safe to assume that every way of building the eventcatalog will result in `eventcatalog.config.js` file being copied next to `next.config.js`. At least thats what I observed during testing this change in this repo and in my local eventcatalog.

### Have you read the [Contributing Guidelines on pull requests]
yes

### DIFF
--- a/.changeset/nine-panthers-hide.md
+++ b/.changeset/nine-panthers-hide.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+Configure base path
--- a/.changeset/wise-carpets-drop.md
+++ b/.changeset/wise-carpets-drop.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+Configure base path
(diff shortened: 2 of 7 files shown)


boyney123:
Thank you @otbe looking good to me, think there might be some lint issues?

`Require statement not part of import statement`

I'm guessing we have to have require here, so maybe ignore this one with inline ignore?

otbe:
Yes @boyney123. If it’s fine for you. I wonder why the typescript rule set lints this file at all but not quite familiar with it. Will add the comment tomorrow (quite late here 😅)

boyney123:
Thank you @otbe.

I made a couple of additions to the pull request, found a couple of broken links (due is the base path is set), mainly around images and schema download. Added small fixes into the PR

boyney123:
@all-contributors please add @otbe for code

PR #120, opened 2022-01-26
feat: Hide owners section if the events has no owners defined

thim81:
## Motivation

The "owners" property on events is optional.

So if the event frontmatter would not have any "owners" defined, you would still see the header & section (see before screenshot). The PR checks if the "owners" are set and if not hides the section.

Before:
2022-01-26 at 14 03 59

After:
2022-01-26 at 14 03 06

(Write your answer here.)
yes

### DIFF
--- a/.changeset/pink-crabs-attend.md
+++ b/.changeset/pink-crabs-attend.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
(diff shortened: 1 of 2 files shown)


boyney123:
Yeah great idea @thim81 , thanks for the pull request, tested it and works great 👍

boyney123:
@all-contributors please add @thim81 for code

PR #123, opened 2022-01-26
feat: asyncapi-plugin Insert external links to the AsyncAPI docs

thim81:
## Motivation

Since EventCatalog generates Events based on the AsyncAPI, it is handy to reference the public AsyncAPI documentation on the Event details page as an external link. This will reduce the workload of adding the links manually.

Result:
2022-01-26 at 15 48 50

Yes

### DIFF
--- a/packages/eventcatalog-plugin-generator-asyncapi/src/__tests__/plugin.spec.ts
+++ b/packages/eventcatalog-plugin-generator-asyncapi/src/__tests__/plugin.spec.ts
@@ -295,7 +295,7 @@ describe('eventcatalog-plugin-generator-asyncapi', () => {
                   - 'Account Service'
               consumers: []
               externalLinks:
-                  - {label: UserSignedUp, href: 'https://eventcatalog.dev/events#message-UserSignedUp'}
+                  - {label: 'View event in AsyncAPI', url: 'https://eventcatalog.dev/events#message-UserSignedUp'}
             ---
 
             <Mermaid />
--- a/packages/eventcatalog-plugin-generator-asyncapi/src/__tests__/plugin.spec.ts
+++ b/packages/eventcatalog-plugin-generator-asyncapi/src/__tests__/plugin.spec.ts
@@ -259,7 +259,7 @@ describe('eventcatalog-plugin-generator-asyncapi', () => {
         it('when includeLinkToAsyncAPIDoc is set, an external link will be added in the event', async () => {
           const options: AsyncAPIPluginOptions = {
             pathToSpec: path.join(__dirname, './assets/valid-asyncapi.yml'),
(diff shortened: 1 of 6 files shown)


boyney123:
Nice one @thim81, thanks for implementing and adding docs and tests!

Just one thought `includeLinkToAsyncAPIDoc` sounds like a boolean vs a string value, I wonder if `externalAsyncAPIUrl` or `externalAsyncAPISpecUrl` something similar would be better suited? What do you think?

website/docs/api/plugins/plugin-doc-generator-asyncapi.md
```diff
@@ -45,6 +45,7 @@ npm install --save @eventcatalog/plugin-doc-generator-asyncapi
 | --- | --- | --- | --- |
 | `pathToSpec` | `string` | `'/asyncapi.yml'` | Path to AsyncAPI document. |
 | `versionEvents` | `boolean` | `true` | When the plugin runs and it finds matching events in the catalog, it will version the events before creating the new documentation. |
+| `includeLinkToAsyncAPIDoc` | `string` | `https://asyncapi.com` | When a AsyncAPI base url is set the, a external link to the AsyncAPI message documentation will be added to each event. |
```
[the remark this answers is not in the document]
thim81: You are correct, I changed it in the last commit

packages/eventcatalog-plugin-generator-asyncapi/src/__tests__/plugin.spec.ts
```diff
@@ -243,6 +246,56 @@ describe('eventcatalog-plugin-generator-asyncapi', () => {
...
+          expect(eventFile).toMatchMarkdown(`
+            ---
+              name: UserSignedUp
+              summary: null
+              version: 1.0.0
+              producers:
+                  - 'Account Service'
+              consumers: []
+              externalLinks:
+                  - {label: 'View event in AsyncAPI', url: 'https://eventcatalog.dev/events#message-UserSignedUp'}
```
thim81, line 298: @boyney123 Are you OK with this syntax? 

```yaml
externalLinks:
                  - {label: 'View event in AsyncAPI', url: 'https://eventcatalog.dev/events#message-UserSignedUp'}
                  ```
                  
 This is how the writeToCatalog injects it. It works since it is a valid syntax, but it looks strange in a YAML file.

boyney123: Yeah think it's OK for now, I was shocked by the format when I saw it first last week, but as you said seems to be valid YML. The lib we use seems to format arrays like this 🤷‍♂️, guess if its a major issue we can fix in the future 👍

PR #139, opened 2022-01-28
feat: configure homepageLink

thim81:
## Motivation

The event catalog can be part of a bigger developer portal. It is a common UX that clicking the logo in the top navigation returns the visitor to the homage page of the website. The configuration is optional, If not set the user will navigate to the root of the Event Catalog.

2022-01-28 at 14 45 13
2022-01-28 at 14 45 49@2x

yes

### DIFF
--- a/examples/basic/eventcatalog.config.js
+++ b/examples/basic/eventcatalog.config.js
@@ -3,7 +3,6 @@ module.exports = {
   tagline: 'Discover, Explore and Document your Event Driven Architectures',
   organizationName: 'Your Company',
   editUrl: 'https://github.com/boyney123/eventcatalog-demo/edit/master',
-  homepageLink: 'https://eventcatalog.dev/',
   trailingSlash: true,
   logo: {
     alt: 'EventCatalog Logo',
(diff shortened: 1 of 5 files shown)


examples/basic/eventcatalog.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   tagline: 'Discover, Explore and Document your Event Driven Architectures',
   organizationName: 'Your Company',
   editUrl: 'https://github.com/boyney123/eventcatalog-demo/edit/master',
+  homepageLink: 'https://eventcatalog.dev/',
```
boyney123, line 6: Maybe for the example (this is what is used for local dev) we should remove this, as I think it would be nice to have local dev copy to go to the `/` (default) route back to the homepage of the Catalog.

boyney123 (changes requested):

boyney123:
Thank you @thim81 looking good! Just a minor change (mentioned above).

examples/basic/eventcatalog.config.js
thim81: It is optional, so if the `homepageLink` is not set, the default will be `/`

thim81: Not sure what I can do.

boyney123: Yeah I think we can remove it from the `examples/basic/eventcatalog.config.js` page

boyney123, line 6:
```suggestion
```

boyney123:
Great thanks again @thim81 !

PR #151, opened 2022-01-31
 feat: json schema viewer for events

thim81:
## Motivation

Event Catalog has a "Viewer" component which fits displaying JSON examples & Code, by using the Prism highlighting.
For the display of JSON schema properties, this PR introduces a dedicated component that shows the properties of the schema like documentation instead of the raw JSON schema.

It is using the Stoplight Json Schema viewer component.
https://www.npmjs.com/package/@stoplight/json-schema-viewer

Story book demo: https://stoplightio.github.io/json-schema-viewer/?path=/story/jsonschemaviewer--default

image

Yes


thim81:
To review how & where to add tests.

Is the styling OK as is or do you want to modify it?

thim81:
@boyney123 Since I needed to add Stoplight dependencies, the Github actions fail on the yarn.lock check. 
Not sure what to do to overcome this?

thim81:
On the styling: The > arrow is now out of proportion.

A simple styling could be added:

```css
.svg-inline--fa.fa-fw {
    width: 1.25em;
}
```

packages/eventcatalog/package.json
```diff
@@ -19,6 +19,10 @@
     "@headlessui/react": "^1.4.2",
     "@heroicons/react": "^1.0.5",
     "@mdx-js/react": "^1.6.22",
+    "@stoplight/json-schema-viewer": "^4.5.0",
```
boyney123, line 22: @thim81 quick question, just looking at their docs, how did you know we need all these dependencies? 

Looking at docs looks like it's just the one, but looks like it doesnt work with just the one either!

packages/eventcatalog/components/SchemaViewer.tsx
```diff
@@ -0,0 +1,33 @@
...
+function SchemaViewer({ schema, renderRootTreeLines = false, hideExamples = false, defaultExpandedDepth = 1 }: Props) {
+  const JsonSchema = useMemo(() => JSON.parse(schema as string), [schema]);
+
+  return (
+    <JsonSchemaViewer
+      schema={JsonSchema}
+      emptyText="No schema defined"
+      defaultExpandedDepth={defaultExpandedDepth}
+      // viewMode='standalone'
+      renderRootTreeLines={Boolean(renderRootTreeLines)}
```
boyney123, line 26: I think we are OK not to cast as boolean, as the types should help here and think it's safe to assume boolean coming into props?
```suggestion
      renderRootTreeLines={renderRootTreeLines}
```

packages/eventcatalog/components/SchemaViewer.tsx
```diff
@@ -0,0 +1,33 @@
...
+      hideExamples={Boolean(hideExamples)}
```
boyney123, line 27:
```suggestion
      hideExamples={hideExamples}
```

packages/eventcatalog/components/SchemaViewer.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { useMemo } from 'react';
```
boyney123, line 1: We have a `components/Mdx` folder for MDX components (although not perfect), wonder if the component should live inside that, considering its a EventCatalog MDX one? What ya think?

packages/eventcatalog/pages/events/[name].tsx
```diff
@@ -46,6 +47,18 @@ export const getComponents = ({ event, schema, examples }: any) => ({
       </section>
     );
   },
+  SchemaViewer: ({ title = 'Event Schema' }: { title: string }) => {
+    if (!schema) return null;
+
+    return (
+      <section className="mt-8 xl:mt-10">
+        <h2 id="activity-title" className="text-lg font-medium text-gray-900 underline">
```
Comment (boyney123) [inline packages/eventcatalog/pages/events/[name].tsx:55]: Think we can remove the `id` here
```suggestion
        <h2 className="text-lg font-medium text-gray-900 underline">
```

packages/eventcatalog/package.json
thim81: When trying to make it work, I had only the `@stoplight/json-schema-viewer` but then the React was complaining about missing packages, so I added them 1 by 1.

packages/eventcatalog/components/SchemaViewer.tsx
thim81: Agreed, I'll move it

thim81: Done

boyney123:
Nice work @thim81 just trying to sort out the builds at the moment.

I also added a new prop `maxHeight` just in case anyone wants to set a height to their component, it's optional thought it would be a nice touch.

boyney123:
Nice! Let's GO!

PR #175, opened 2022-02-04
fix: mermaid service publish - subscribe

thim81:
## Motivation

When the same events/services are being used in different directions in a diagram, Mermaid considers them the same and renders them as 1 node. This results in graphs like in the screenshot. Not sure if this is desired behaviour?

The PR solves it by making the nodes unique per direction, by adding a little prefix in front of the ID.

Before PR:
2022-02-04 at 23 44 45
2022-02-04 at 23 56 57

After PR: 
2022-02-04 at 23 53 09
2022-02-04 at 23 57 29

(Write your answer here.)

### DIFF
--- a/.changeset/popular-apes-heal.md
+++ b/.changeset/popular-apes-heal.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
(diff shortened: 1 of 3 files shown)


boyney123:
Thank you @thim81 looks loads better!

PR #184, opened 2022-02-07
fix: increase node width for NodeGraph

thim81:
## Motivation

The NodeGraph minimum width logic resulted in a too small node.
This PR increased the node width for NodeGraph.

before:
2022-02-07 at 20 53 01@2x

after:
2022-02-07 at 20 51 59@2x

2022-02-07 at 20 49 29@2x

Yes

### DIFF
--- a/.changeset/khaki-colts-appear.md
+++ b/.changeset/khaki-colts-appear.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+fix: increase node width for NodeGraph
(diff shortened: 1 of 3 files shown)


boyney123:
Thanks @thim81 👍

boyney123:
Let's give this ago, my suspension is we may fall into the same trap with other words, let's try.

thim81:
In the logic, I bumped the multiplier but kept the minimum width, so that the edge cases should at least occur less.

PR #207, opened 2022-02-28
allow deep link to an event field

donaldpipowitch:
## Motivation

In order to share documentation for single fields in long event definitions efficiently it would be nice, if we could use individual fields as a target.

Ideally this feature would be build into `@stoplight/json-schema-viewer`, but thanks to `renderRowAddon` it can also be implemented in user land.

Sadly because we collapse event definitions by default it is not super easy to target them. We actually have to wait for the client side rendering to become available to check, if an anchor was used. Due to some rendering bug we then have to re-render `<JsonSchemaViewer/>` by changing the `key` (this of course only happens if you make use of this feature - no negative impact for other users). After that we can manually scroll to the field.

image

(Added the small `#` icon at the end.)

Yes

### DIFF
--- a/.changeset/late-tips-lay.md
+++ b/.changeset/late-tips-lay.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+allow deep linking event field
--- a/packages/eventcatalog/components/Mdx/SchemaViewer/SchemaViewer.tsx
+++ b/packages/eventcatalog/components/Mdx/SchemaViewer/SchemaViewer.tsx
@@ -1,11 +1,14 @@
-import React, { useMemo } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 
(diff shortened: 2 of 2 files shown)


boyney123:
Thanks for this improvement @donaldpipowitch 

Do you think it's worth adding this feature to the docs? https://www.eventcatalog.dev/docs/components/overview#schemaviewer-

Just an overview of what it is and how you can use it?

donaldpipowitch:
As a regular feature I'd personally  not need it in the docs. (I'd hope it's self explanatory, similar to the anchors we see in GitHub when hovering over a heading in a markdown file.)

But... if you want this to be a configurable feature I would definitely add it there. Should I make it configurable?

boyney123:
Hey @donaldpipowitch , sorry for the late review here, think EventCatalog has evolved since these PR were raised. Going to close for now, if you need this feature, feel free to raise again and I will take a look ASAP. Sorry it's taken so long!

PR #241, opened 2022-03-16
improve typings

donaldpipowitch:
## Motivation

As [mentioned on discord](https://discord.com/channels/918092420338569216/918092420338569219/948603441528328192) I tried to add a couple of bugfixes which was hindered by current typings.

In this MR I did three things:
1. update TypeScript
2. get rid of `[]` usage in array (e.g. `producerNames?: string[] | Service[] | [];`)
3. introduce union type (for `data: A | B;  source: 'a' | 'b';`)

About **2.**: A type like `string[] | Service[]` already implies that the array can be empty. `const a: string[] | Service[] = []` is completely fine. The type `[]` can actually be worse as it says _you can only be empty_. E.g. mapping over such an collection uses never `empty.map(/** never **/ item => item)`.

About **3*.*: Switching from `data: A | B;  source: 'a' | 'b';` to `{ data: A;  source: 'a'; } | { data: B;  source: 'b'; }` makes checks on `source` and further usage of `Data` type safe: `if (source === 'b') // Data has the type B`.

Look at this example:
image

By checking `source === 'domain' || source === 'all'` TypeScript knows the type of `data`. We don't need custom `as` usage anymore.

Yes.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
+}
--- a/.changeset/olive-dingos-reflect.md
+++ b/.changeset/olive-dingos-reflect.md
@@ -1,0 +1,7 @@
+---
+"@eventcatalog/types": patch
+"@eventcatalog/utils": patch
+"@eventcatalog/core": patch
+---
+
+improve typings
--- a/.changeset/olive-dingos-reflect.md
+++ b/.changeset/olive-dingos-reflect.md
@@ -4,4 +4,4 @@
 "@eventcatalog/core": patch
 ---
 
-improve typings
+chore: improve typings
--- a/package.json
+++ b/package.json
@@ -57,6 +57,6 @@
     "lerna": "^4.0.0",
     "lerna-changelog": "^2.2.0",
     "prettier": "^2.1.2",
-    "typescript": "^4.5.2"
+    "typescript": "^4.6.2"
   }
 }
--- a/packages/eventcatalog-types/src/index.d.ts
+++ b/packages/eventcatalog-types/src/index.d.ts
@@ -22,13 +22,13 @@ export interface Event {
   draft?: boolean;
   summary?: string;
   domain?: string | null;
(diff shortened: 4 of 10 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
```
boyney123, line 2: Hey @donaldpipowitch ,

Any info on what this vscode setting does?

boyney123:
Hey @donaldpipowitch,

Great work! Thanks for the improvements 🙏,

I just had a small comment on the vscode settings, apart from that looking good to me !

.vscode/settings.json
donaldpipowitch: Hi @boyney123,

sorry, this change accidentally slipped in. I can revert it, if you want.
But in case you use VS Code... this one is actually quite handy, I use it basically on every of my own projects. By setting `"typescript.tsdk"` you tell VS Code which TypeScript version should be used for the Language Server. It is either the one build into VS Code or the one belonging to the project. By not relying on the VS Code internal TypeScript version (which could be newer or older then the TypeScript version referenced in your project) you should always see the same errors in your editor as you see it when you type check your project via the terminal.

boyney123: Cool thanks for explaining 👍 

Yeah let's keep it in then, sounds good to me!

boyney123:
@donaldpipowitch just checked out the branch, and my VSCode giving me some type errors around some changes you did?

image

Do you have the same?

donaldpipowitch:
@boyney123 I bet you have to restart the TypeScript server (or close and open VS Code again).
image

This can happen when the underlying TypeScript version changes when you switch branches/install new TypeScript versions. (Have you run `$ yarn` to install TypeScript 4.6?)

boyney123:
Cool thanks @donaldpipowitch your stuff fixed the issue! Great I will get this released!
Thanks so much for the contribution 🙏

boyney123:
@allcontributors add @donaldpipowitch for code

PR #248, opened 2022-04-07
fix(core): linking to event in domain graphs

donaldpipowitch:
## Motivation

Fix https://github.com/boyney123/eventcatalog/issues/223

Yes.

---

This MR fixes https://github.com/boyney123/eventcatalog/issues/223. I'm not sure if more places need to be fixed however. 

E.g. this part:

```
if (source === 'domain' || source === 'all') {
      const totalEventElements = data.events.map((event) =>
        getEventElements({
          event,
          rootNodeColor,
          isAnimated,
          includeLabels: true,
          includeNodeIcons: true,
          source: source === 'domain' ? { type: 'domain', name: data.name } : undefined,
        })
      );
```

I'm not sure when `source` can be `'all'`. For now I passed `source` only to `getEventElements` when it's `'domain'`.

---

FYI: If you move comments for params out of the JSDoc comment into the type signature they are actually shown on hover in VS Code.

image

### DIFF
--- a/packages/eventcatalog/components/Mdx/NodeGraph/GraphElements.tsx
+++ b/packages/eventcatalog/components/Mdx/NodeGraph/GraphElements.tsx
@@ -8,7 +8,20 @@ const { publicRuntimeConfig: { basePath = '' } = {} } = getConfig();
 const MIN_NODE_WIDTH = 150;
 type NODE_TYPES = 'service' | 'event';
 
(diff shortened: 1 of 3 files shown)


boyney123:
Hey @donaldpipowitch , sorry it's taken so long to get back to you, think we had another PR very simular that also fixes other places with wrong urls, https://github.com/boyney123/eventcatalog/pull/261

I'm going to close this one, as I believe when the other one is merged it should all be fixed!

Thank you for the patience and also the PR, looks like EventCatalog is one step closer to supporting these nested links thanks to you all!

PR #249, opened 2022-04-09
fix(core): show non-latest consumers and producers

donaldpipowitch:
## Motivation

Fix https://github.com/boyney123/eventcatalog/issues/85.

Yes.

---

This is my first stab at fixing https://github.com/boyney123/eventcatalog/issues/85. I'd like to know, if this is already enough for the first iteration and if not what is missing.

Two things I noticed:

**1.**: The existing `getAllEventsAndVersionsFlattened` seems to be buggy (excluding the most recent versions, because of a line which was commented out) and not returning full events (just som fields). That's why I made `getAllEvents` to allow returning versions as well.

**2.**: I stumbled over several problems, because `noImplicitAny` is not set in the `tsconfig.json`. Am I allowed to introduce this in another PR? For example the whole implementation of `truncateNode` in `graphs.ts` didn't really made any sense currently (see my fixed version). I'm also wondering if it was actually used correctly (`centerNode` used it for `id` so I did the same for `leftNodes` and `rightNodes`, but I'd assume you want it for `name`...?).

### DIFF
--- a/examples/basic/domains/Shopping/events/AddedItemToCart/index.md
+++ b/examples/basic/domains/Shopping/events/AddedItemToCart/index.md
@@ -1,6 +1,6 @@
 ---
 name: AddedItemToCart
-version: 0.0.2
+version: 0.0.3
 summary: |
   Holds information about what the user added to their shopping cart.
 producers:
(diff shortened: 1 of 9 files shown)


boyney123:
Hey @donaldpipowitch , sorry for the late review here, think EventCatalog has evolved since these PR were raised. Going to close for now, if you need this feature, feel free to raise again and I will take a look ASAP. Sorry it's taken so long!

PR #256, opened 2022-04-22
fix(core): Fixes node domain service/event linking

### LINKED ISSUE #255: Links in NodeGraphs do not take domains into account
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/boyney123/eventcatalog/blob/master/CONTRIBUTING.md#reporting-new-issues).

### Description

Clicking on a node in a NodeGraph that shows services and events that belong to domains links to top-level `services` and `events` directories rather than `domains/<domain>/services` and `domains/<domain>/events` directories.

### Steps to reproduce

1. Create a domain
2. Add a service that produces an event
3. Click on the Visualizer
4. Click on the event created in step 2 that belongs to the domain
5. Get a 404 because it goes to `/events/<event-name>` rather than `/domains/<domain>/events/<event-name>`

### Expected behavior

Should navigate to the `/domains/<domain>/events/<event-name>` documentation.

### Actual behavior

Navigates to the `/events/<event-name>` resulting in a 404 not found.

### Your environment

- EventCatalog version used: 0.2.14

drub0y:
Fixes #255

## Motivation

Node linking in graphs was broken for domain based services and events. Want those links to work! 😊

Yep! 👍


boyney123:
Thanks for the PR @drub0y, just reviewed it and mostly seems to be working OK :)

There is one slight issue I found though, it's when we have nodes in the node graph link to other things outside it's domain.

To give an example:

Here is the `OrderRequested` event (https://app.eventcatalog.dev/domains/Orders/events/OrderRequested/)

The **OrderRequested** event is in the **Orders** domain and the **Payment Service** is in **no domain** but the code assume it's the `orders` domain, which leads to a broken link 

You can see this if you run the code in the PR and go to `OrderRequested` and click on the `Payment Service` node, which breaks (as the `orders) domain is assumed.`

Hmm it's tricky as we only have the names for the producers/consumers and would be better if we actually had the objects so we can pass the domains that they belong too... I will explore to see what we can do, but I remember this being a design issue

boyney123:
Reviewed the code and added a few more things as I found the issue above, going to close this one in favor of the other (which is simular just fixed bugs on top).

https://github.com/boyney123/eventcatalog/pull/261

PR #261, opened 2022-05-03
fix(core): Fixes node domain service/event linking

boyney123:
## Motivation

Extension to #256


boyney123:
Think this is good to go @drub0y , once you are cool with it we can get it merged and released :)

drub0y:
Re-tested various linking scenarios to both domain based events/services and non-domain based and it seems like it's all working to me! 👍 :shipit:

boyney123:
Thanks @drub0y ! Shipping now

PR #277, opened 2022-05-25
feat(utils/generator-asyncapi): Add domain parsing

dpwdec:
## Motivation

Eventcatalog does not currently support the interop of AsyncAPI documents into domain boundaries. This functionality is also not supported directly by the AsyncAPI spec and so it makes sense to include the domain boundaries as part of the eventcatalog config rather than trying to develop a new standard for domain boundaries using `x-properties` in the AsyncAPI document. 

To this end, if users would like to break up their AsyncAPI document parsing by domain then each generator in the `generators` field of the `eventcatalog.config.js` file can map to each domain required. If domains are not included then the generator works as before and pushes services and events into the "global" catalog.

A usage example for this would be:
```js
// eventcatalog.config.js
const path = require('path');

module.exports = {
   generators: [
    // first generator pushes parsed files into the Orders domain
    [
      '@eventcatalog/plugin-doc-generator-asyncapi',
      {
        // path to your AsyncAPI files
        pathToSpec: [path.join(__dirname, 'asyncapi-order-created.yml'), path.join(__dirname, 'asyncapi-order-deleted.yml')],

        // version events if already in catalog (optional)
        versionEvents: true,
        domainName: 'Orders'
      },
    ],
    // second generator pushes parsed files in the Payments domain
    [
      '@eventcatalog/plugin-doc-generator-asyncapi',
      {
        // path to your AsyncAPI files
        pathToSpec: [path.join(__dirname, 'asyncapi-payments-processed.yml')],

        // version events if already in catalog (optional)
        versionEvents: true,
        domainName: 'Payments'
      },
    ],
  ],
};
```

Yes 🐒

### DIFF
--- a/.changeset/rotten-numbers-scream.md
+++ b/.changeset/rotten-numbers-scream.md
@@ -1,5 +1,6 @@
 ---
 '@eventcatalog/utils': minor
+'@eventcatalog/plugin-doc-generator-asyncapi': minor
 ---
 
 feat(utils): Add utilities for writing and reading domains to the catalog
--- a/.changeset/rotten-numbers-scream.md
+++ b/.changeset/rotten-numbers-scream.md
@@ -1,0 +1,5 @@
+---
+'@eventcatalog/utils': minor
+---
+
+feat(utils): Add utilities for writing and reading domains to the catalog
--- a/.changeset/tall-dogs-pretend.md
+++ b/.changeset/tall-dogs-pretend.md
@@ -1,5 +1,0 @@
----
-'@eventcatalog/plugin-doc-generator-asyncapi': minor
----
-
-feat(generator-asyncapi): Add options for parsing AsyncAPI files into domains
--- a/.changeset/tall-dogs-pretend.md
+++ b/.changeset/tall-dogs-pretend.md
@@ -1,0 +1,5 @@
+---
+'@eventcatalog/plugin-doc-generator-asyncapi': minor
+---
+
+feat(generator-asyncapi): Add options for parsing AsyncAPI files into domains
--- a/packages/eventcatalog-plugin-generator-asyncapi/src/__tests__/plugin.spec.ts
+++ b/packages/eventcatalog-plugin-generator-asyncapi/src/__tests__/plugin.spec.ts
@@ -418,10 +418,12 @@ describe('eventcatalog-plugin-generator-asyncapi', () => {
           // just wait for files to be there in time.
           await new Promise((r) => setTimeout(r, 200));
 
+          const { getDomainFromCatalog } = utils({ catalogDirectory: process.env.PROJECT_DIR });
           const { getEventFromCatalog, getServiceFromCatalog } = utils({ catalogDirectory: path.join(process.env.PROJECT_DIR, 'domains', options.domainName) });
 
           const { raw: eventFile } = getEventFromCatalog('UserSignedUp');
           const { raw: serviceFile } = getServiceFromCatalog('Account Service');
+          const { raw: domainFile } = getDomainFromCatalog('My Domain');
 
           expect(eventFile).toMatchMarkdown(`
             ---
@@ -446,6 +448,15 @@ describe('eventcatalog-plugin-generator-asyncapi', () => {
 
             <NodeGraph />`
           );
+
+          expect(domainFile).toMatchMarkdown(`
+            ---
+            name: 'My Domain'
+            summary: 'A summary of my domain.'
+            ---
+
+            <NodeGraph />
+          `);
         });
       });
     });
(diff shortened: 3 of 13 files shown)


mikaelvesavuori:
Regarding: `This functionality is also not supported directly by the AsyncAPI spec and so it makes sense to include the domain boundaries as part of the eventcatalog config rather than trying to develop a new standard for domain boundaries using x-properties in the AsyncAPI document.`

I was actually mentally going down the route of `x-properties` here. For what we want:

* There should be complete automation of all typical data and fields (like "domain") from schemas (which teams own), not from EventCatalog config (which they don't own/control)
* EventCatalog should be a single enterprise-facing service that collects all services, not on individual "per-service" level, which logically means we should trust the generator to fully and comprehensively do its job versus also making custom scripts to make dynamic configuration changes or to do manual labor on those files

To get there, I personally believe the schema property makes more sense, or at least to provide the option.

Hope this gets merged, as it is still an improvement though!

dpwdec:
@mikaelvesavuori Great points, totally agree that it should be a "single enterprise-facing service" and I agree with the points on `domainName` being a pain to configure in pure AsyncAPI.

From my perspective, we run a "pre-generate" step during our eventcatalog build process that templates the `eventcatalog.config.js` file based on the domain boundaries that teams configure. So, instead of domain boundaries living inside our AsyncAPI files they exist in a separate org repository that is specifically designed for holding metadata around services and their connections. This metadata is served as an API that we call during the build process to populate things like `domainName` and the correct domain object during the `eventcatalog.congig.js` templating. This is for 500~ microservices, so it's definitely on the upper end of this sort of stuff where we can afford this separation of concerns. But, even with that in mind I would be slightly wary of introducing domains as a concern directly into AsyncAPI based on what is currently supported.

My main problem with using `x-properties` or schema-inferred domains is that we are essentially pushing a concept that exists in eventcatalog (or other domain systems) into the AsyncAPI spec itself which shifts a context into the documentation objects that hasn't been "designed" for it, as it were. Teams who own AsyncAPI then have to understand that their stuff is going to be used by domains or by EC when they document?

That's the reasoning I had for going with the solution in the PR. I'm sure we could have adopted another solution if it existed, or is implemented in the future. I'm not wedded to this and very happy to change or migrate away from it if there's a properly nailed down proposal for another method.

That said, I think this PR is a bit of just an "easy fix" which skirts the issue of how we agree the correct way to use schemas or x-properties to define domains. Which is definitely a design discussion I'm keen to have as a more long term solution and would be better for the health of eventcatalog.

Really do appreciate the feedback and engagement on this! Thanks! 😊

dpwdec:
@boyney123 The PR has been open for 30 days plus now which has given people some time to look at the design and it seems like there aren't big objections to the design thus far, if there's anything else you would like me to change / improve please let me know and I can work on it? Otherwise, is it possible we can move toward merging it so we can start making use of it in releases? Thank you.

mikaelvesavuori:
Thanks for the detailed response! Just a quickie on `Teams who own AsyncAPI then have to understand that their stuff is going to be used by domains or by EC when they document?` In our organization, we have a relatively stable domain-oriented foundation, and I believe anyone working in such an org (or with domain driven design generally) is bound to have to expose semantics around "domains" to the actual implementing teams. However, as you write, it is not an ideal mechanism to place it in the specification—I'd be really happy if this was a first-class property though!

We have an even bigger estate than you, yet a poorer overall mapping, and my hope is that EventCatalog could become that place, ergo my direction on this topic.

Again, thanks for the work on an improvement to the current state anyway!

boyney123:
Thanks the patience @dpwdec, planning to look at this ASAP

boyney123:
Hey @dpwdec ,

Just reviewed, and have to say this is a great pull request, solution is spot on and even adding the docs 😍,

Sorry taken me so long to review, I'm just going to make a release now and will post back here with the new versions.

I'm also looking for other contributors to help with the project, so if you ever fancy it, let me know! 

Thanks again, this is a great feature.

boyney123:
@all-contributors add @dpwdec for docs,code

boyney123:
https://github.com/all-contributors?type=source add @dpwdec for doc code

boyney123:
@all-contributors add @dpwdec for doc code

boyney123:
New version is out for EventCatalog! 

AsyncAPI users can now use the plugin to group there services and events with **domains**!

To update
    - `@eventcatalog/utils@0.2.0`
   - `@eventcatalog/plugin-doc-generator-asyncapi@0.3.0`

Read more on the docs:

https://www.eventcatalog.dev/docs/api/plugins/@eventcatalog/plugin-doc-generator-asyncapi#what-if-i-want-to-parse-asyncapi-files-into-different-domains

Thanks for the contribution!

PR #320, opened 2022-11-04
Mermaid component type error breaks build

mikaelvesavuori:
Hi! The `width` property usage here will break the build as the types are not compatible/expecting these keys. Removing them makes it all work and build as expected.


boyney123:
Thanks @mikaelvesavuori , looks like mermaid release changes the types, just visually checked and seems to render the same without the widths set too, thanks for the PR and fix!

PR #350, opened 2023-02-22
fix: correct display of users in domain

ldiego73:
## Motivation

The following PR solves these two problems:
1. when I create a domain without events and add a domain owner, when clicked it does not display returning 404 not found.
2. The user page only returns the information (domains, events and services) that counts available.

yes


boyney123:
Thank you @ldiego73 , just tested this and replicated the issue, your fix works well! Thank you

boyney123:
@all-contributors add @ldiego73 for code

Issue #387, opened 2023-06-02
Expose more OpenAPI component options


kernwig:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/boyney123/eventcatalog/blob/master/CONTRIBUTING.md#reporting-new-issues).

### Description

The default output of the <OpenAPI/> (SwaggerUI under the hood) is to require clicking on every property individually to expand it. I find this exasperating; an entire level should expand at once.

[SwaggerUI](https://www.npmjs.com/package/swagger-ui-react) has options to control this, they just need to be exposed: `docExpansion` and `defaultModelExpandDepth`.

The options `defaultModelRendering` and `displayOperationId` also look useful.

(I'd offer a PR, but simply `git clone` and `npm install` fail for me.)

### Motivation

Allow for further customization to project needs.

boyney123:
Thanks @kernwig , I have made a feature change to allow this, thanks for the idea. Going live now.

kernwig:
Thank you. :-) I ended up using 

`<OpenAPI showTitle="false" defaultModelExpandDepth="2" defaultModelRendering="model"/>`

Appreciate the Swagger UI update as well.

Issue #378, opened 2023-04-07
OpenAPI/AsyncAPI from remote URL


thawankeane:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/boyney123/eventcatalog/blob/master/CONTRIBUTING.md#reporting-new-issues).

### Description

Currently, the only way to [add a service OpenAPI](https://www.eventcatalog.dev/docs/services/adding-service-openapi) (AyncAPI also) is to add the spec file into the service directory, but doing this can be painful to keep it up sync between the projects.

It will be nice if we could configure one or more spec URLs inside the [service frontmatter configuration](https://www.eventcatalog.dev/docs/api/event-frontmatter) alternatively to the physical files.

This is my proposal (if it's possible):

```md PetstoreService/index.md
---
name: PetstoreService
summary: |
  Lorem ipsum dolor sit amet, consectetur adipisicing elit. Beatae in ex quaerat cum nihil expedita recusandae non, dolores corporis corrupti odio reprehenderit repellat possimus aliquid vero exercitationem!
owners:
  - johndoe
openapi:
  - https://petstore3.swagger.io/api/v3/openapi.json
---

Lorem ipsum dolor sit amet consectetur adipisicing elit. Consequatur unde adipisci vel. Perferendis, inventore quas maxime a dolores consequatur soluta iste nisi, exercitationem beatae quidem ad. Velit voluptatem veniam libero.

<OpenAPI />

<NodeGraph />
```

### Motivation

Many times, OpenAPI/AsyncAPI specifications are hosted into remote URLs, it easier to keep everything synced. Currently, to make eventcatalog updated with those remote URLs, developers (or CI scripts) must download these files from the remote and put them in each service directory.

boyney123:
Hey @thawankeane 

Just released this one https://github.com/boyney123/eventcatalog/issues/387

You should be able to pass any probs to this component now, I believe if you pass the `url` prop to the component with the url of the file it will work.

Closing this issue for now, I tested locally and worked :)

PR #428, opened 2023-08-26
feat: Allow fetching of remote asyncapi docs

jlee-spt:
## Motivation

Related to #378 and #387. This PR allows the AsyncAPI docs to be fetched from a remote URL by exposing a new AsyncAPI component option (url). Under-the-hood, it uses [FetchingSchemaInterface](https://github.com/asyncapi/asyncapi-react/blob/master/library/src/types.ts#L396) when the url option is specified.

Yes

### DIFF
--- a/packages/eventcatalog/pages/services/[name].tsx
+++ b/packages/eventcatalog/pages/services/[name].tsx
@@ -37,8 +37,8 @@ function MermaidComponent({ title, service, charts }: { title?: string; service:
 const getComponents = (service: Service) => ({
   Admonition,
   AsyncAPI: ({ url }: { url?: string }) => {
-    if(url) return <AsyncApiSpec url={url} />
-    
+    if (url) return <AsyncApiSpec url={url} />;
+
     if (!service.asyncAPISpec && !url) return null;
     return <AsyncApiSpec spec={service.asyncAPISpec} />;
   },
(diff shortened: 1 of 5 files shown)


jsteiml:
This would be a game changer for us. Currently if anything changes you have to manually go in and update the the spec for each event (and we have quite a few) This would remove that step for us.

packages/eventcatalog/pages/services/[name].tsx
```diff
@@ -36,9 +36,9 @@ function MermaidComponent({ title, service, charts }: { title?: string; service:
 
 const getComponents = (service: Service) => ({
   Admonition,
-  AsyncAPI: () => {
+  AsyncAPI: ({ url }: { url?: string }) => {
     if (!service.asyncAPISpec) return null;
-    return <AsyncApiSpec spec={service.asyncAPISpec} />;
+    return <AsyncApiSpec spec={service.asyncAPISpec} url={url} />;
```
Comment (boyney123) [inline packages/eventcatalog/pages/services/[name].tsx:41]: Not sure if I'm missing something, but is this valid? https://github.com/asyncapi/asyncapi-react#props

I can't see a URL prop for this component? Can you link some docs?

Comment (boyney123) [reply packages/eventcatalog/pages/services/[name].tsx:41]: Sorry think I found it now FetchingSchemaInterface, thanks looks good to me, will try it out....

Comment (boyney123) [reply packages/eventcatalog/pages/services/[name].tsx:41]: Sorry also helps if I read your PR description, sorry 😅

boyney123:
@all-contributors add @jlee-spt for code

PR #440, opened 2023-09-13
Added users page

rtoro:
## Motivation

It is useful to have a page to list all the users configured in the catalog
Currently you can only review users one at a time, with this change the /users page is enabled

yes


boyney123:
Thank you @rtoro , think this is a great feature to add. I edited some styles slightly, and also added the new page to the header nav in new catalog config files.

Issue #484, opened 2024-02-16
Clicking a non Event/Service graph node results in a runtime error


dreglad:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/boyney123/eventcatalog/blob/master/CONTRIBUTING.md#reporting-new-issues).

### Description

In the visualizer graph, when clicking on a publishes/subsribed-by node label, a runtime error is thrown because of the `onElement` callback is trying to read a property from an undefined argument prop

### Steps to reproduce

1. Create new project with `npx @eventcatalog/create-eventcatalog@lates`
2. Run and Navigate to a service in the visualizer  (e.g. http://localhost:3000/visualiser/?type=event&name=AddedItemToCart)
3. Click on any gray `subscribed by` or `publishes` label

### Expected behavior

No error is thrown

### Actual behavior

<img width="1307" alt="screen" src="https://github.com/boyney123/eventcatalog/assets/50302/6493830a-734a-440e-8af5-43b99717dd2a">

### Your environment

- EventCatalog version: current default branch head (fda490cc15770ad947b4b423e127877f8261a4f1)

boyney123:
Thanks for raising this issue. 

EventCatalog v2 is coming out very soon, so going to close this as I believe it’s no longer an issue in the new version. 

If you still experience this issue please raise a new issue. 

Thank you!

PR #485, opened 2024-02-16
fix: (#484) ensure link exists before opening link in visualizer

dreglad:
## Motivation

Fixing bug described in #484

Adding a conditional to the `ReactFlow`'s `onElementClick` callback function to ensure a link actually exists before navigating to it


boyney123:
Thanks @dreglad makes sense to me!

boyney123:
@all-contributors add @dreglad for code

Issue #498, opened 2024-04-03
Search directly via the URL


marciovmartins:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/boyney123/eventcatalog/blob/master/CONTRIBUTING.md#reporting-new-issues).

### Description

Option to get a term searched via the URL, for example:
- https://app.eventcatalog.dev/events?q=Order
- https://app.eventcatalog.dev/services?q=Order
- https://app.eventcatalog.dev/domains?q=Order

This would already load the page with the "Search Events|Service|Domain" filled and the result filtered.

### Motivation

This would enable third-party tools to have shortcuts to directly search for the "Event" like CLI alias or Alfredapp.

```sh
# could open https://app.eventcatalog.dev/events?q=Order
sre Order
# could open https://app.eventcatalog.dev/services?q=Order
srs Order
# could open https://app.eventcatalog.dev/domains?q=Order
srd Order
```

Instead of open the Event Catalog in the web browser, click in the "Search Events" input box, fill the text and wait for the result to be queried.

boyney123:
Yeah like the idea!

boyney123:
EventCatalog will now have support for search in v2, will close this for now. Let me know if you still want this support or not in v2.

PR #499, opened 2024-04-04
feat: add search via query parameter for events, services and domains (#498)

marciovmartins:
## Motivation

Implementing suggestion described here (#498)

Changed the input to be a react managed component and checking if the query string "q" is provided.

Can be tested locally calling http://localhost:3000/events/?q=Order and http://localhost:3000/services/?q=Order and http://localhost:3000/domains/?q=Order 

The expected behaviour is that the result list be filtered by the term in the query parameter "q".

Let me know if you have a different suggestion for query parameter.

### DIFF
--- a/.changeset/cyan-zebras-happen.md
+++ b/.changeset/cyan-zebras-happen.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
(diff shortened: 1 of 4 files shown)


boyney123:
Thank you @marciovmartins !

PR #515, opened 2024-04-30
feat(services): support multiple markdown files

gjuchault:
## Motivation

This closes https://github.com/boyney123/eventcatalog/issues/242

The goal of this PR is to introduce "extra files" — documentation. I think this could be valuable to have files from "repository starter" like CONTRIBUTING.md, or more domain-oriented files like feature descriptions, etc.

image
image

Some points that can be discussed:
- I included a commit to set a volta pin in the package.json to ease development phase (but I can drop this from the PR if you want)
- I included a commit to fix the `start-catalog-locally` command that was failing
- Should this PR also support nested files? I think this could bring visual complexity (how to display them in the sidebar?)

### Have you read the [Contributing Guidelines on pull requests]

Yes

### DIFF
--- a/.changeset/two-sloths-type.md
+++ b/.changeset/two-sloths-type.md
@@ -1,0 +1,6 @@
+---
+'@eventcatalog/types': patch
+'@eventcatalog/core': patch
+---
+
+support multiple markdown files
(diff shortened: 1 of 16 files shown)


anaelChardan:
Would be awesome to have this one!

boyney123:
Closing for now, as EC V2 now supports custom components which will support this.

Thanks for the idea and PR @gjuchault 

https://www.eventcatalog.dev/docs/development/guides/custom-components/introduction

Issue #541, opened 2024-07-05
Building EventCatalog does not seem to include content


Laupetin:
Running `npm run build` with a new demo event catalog setup does not seem to include content like services,events,domains,commands,...

Looking at the build log, the following is logged:
![image](https://github.com/event-catalog/eventcatalog/assets/9197140/78853514-4248-4de1-a9dc-8642e91f566f)

Running `npm run dev` seems to work fine however.

boyney123:
Thanks! And thanks for the PR

Fixed in @eventcatalog/core@2.0.6

PR #542, opened 2024-07-05
fix(core): hydrate content before building

Laupetin:
fixes: #541 

I have no idea whether this is a good solution or not.
It seems to fix the problem for me.

If there is a better solution feel free to scrap this :)


boyney123:
Ah yeah makes sense to me, the files copy over to the catalog under the hood. This was missing.

Thanks!

boyney123:
@all-contributors add @Laupetin for code

Issue #560, opened 2024-07-08
Multiline summary is displayed on a single line in the visualiser


debae:
### Summary
Multiline summary is displayed on a single line in the visualiser

#### Description
When using multiline markdown like in the example below
```
---
id: cancel-transportation-order-command-failure-response
name: CancelTransportationOrderCommandFailureResponse
version: 6.0.0
summary: |
  When Node Orderplanning commands an Transportation Order Cancel, 
  this topic is provided in the command, so logistics (Transportation Planning) 
  cap knows to produce the response to this topic
```
this still gets displayed as a single line in the visualiser like in the screenshot below :

![image](https://github.com/event-catalog/eventcatalog/assets/2007116/459304a5-7821-49da-ad67-c301156db72b)

boyney123:
Thank you! And also thanks for fixing.

Changes are now in `@eventcatalog/core@2.0.9`

PR #562, opened 2024-07-09
fix(core): configure width for nodes in visualiser (#560)

debae:
* configure min-width and max-width for events/service/command nodes in visualise


debae:
With the suggested changes, the visualiser will now look like this when adding long text.

image

boyney123:
Thank you @debae , LGTM, just tried it out works fine

boyney123:
@all-contributors add @debae for code

PR #593, opened 2024-07-12
fix(core): fixed typescript issues

### LINKED ISSUE #552: eventcatalog build fails with TS error if you have many objects
# Summary

When trying to build the catalog via `eventcatalog build` it fails with the following TS error:

```
src/components/DocsNavigation.astro:19:17 - error ts(2590): Expression produces a union type that is too complex to represent.

19 const allData = [ ...domains, ...services, ...messages, ...teams, ...users];
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

We work with 30 domains, 25 services and ~950 events. dev mode wokrs fine.

otbe:
closes #552

### DIFF
--- a/.changeset/tough-spiders-repair.md
+++ b/.changeset/tough-spiders-repair.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+fix(core): fixed typescript errors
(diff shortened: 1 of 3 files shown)


boyney123:
@all-contributors add @otbe for code

PR #596, opened 2024-07-16
fix: Update Discover nav layout

### LINKED ISSUE #595: Domains show as active even if there are no domains
When loading the Discover page, the Domains link is active, even if there are 0 domain. Clicking it then returns a 404 because there are no domains.

image

ryancormack:
Use domains count, rather than service count when deciding to enable Domains link

## Motivation

Fixes #595


boyney123:
@all-contributors add @ryancormack for code

boyney123:
Thanks @ryancormack

PR #597, opened 2024-07-16
feat(config): add port config

nathanbirrell:
## Motivation

Our team already runs Docusaurus on port 3000 and we'd like to run the eventcatalog server simultaneously locally. Having a configurable port would be helpful in this case. 

## To do

- [ ] update the docs — (Question: where is the docs repository? I can't find it in the repo, and the "Edit this page" links are broken)
- [ ] consider raising a ticket for allowing complete configurability for Astro's config. ie spread an "astroConfig" onto a default base of settings.

### DIFF
--- a/.changeset/brown-wolves-judge.md
+++ b/.changeset/brown-wolves-judge.md
@@ -1,0 +1,5 @@
+---
+'@eventcatalog/core': patch
+---
+
+feat(config): add port config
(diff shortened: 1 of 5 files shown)


boyney123:
Thanks for the PR @nathanbirrell, nice addition.

The docs at the moment are still private, I need to work on making that public, on the list, a few I need to change on the website before I can do this, so I will add the docs no problem.

In terms of spreading Astro, I'm in two minds here. I wanted to keep astro and EventCatalog users split really (well that was the idea...), so the config is seperate from astro, but things naturally don't seem to be going that way, so if we have loads more in the future, I might just allow the spread.... let's see!

boyney123:
@all-contributors add @nathanbirrell for code

PR #611, opened 2024-07-19
fix(core): fixed favicon url

Laupetin:
## Motivation

When using the default base url configuration it defaults to `/`.
The favicon url is currently evaluated from 
```astro
href={`${import.meta.env.BASE_URL}/favicon.ico`}
```
which results in a url of `//favicon.ico`.

```html
<link rel="icon" type="image/svg+xml" href="//favicon.ico">
```

The browser interprets that as a host and tries to load a favicon from `https://favicon.ico/` (if the catalog url uses https as protocol).

This PR makes the favicon url use the utils method that other places use as well to calculate its url.

### DIFF
--- a/.changeset/five-terms-try.md
+++ b/.changeset/five-terms-try.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+fix(core): favicon having invalid url with default base url configuration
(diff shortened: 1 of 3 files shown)


boyney123:
Nice find, thank you @Laupetin 🙏

boyney123:
@all-contributors add @Laupetin for code

PR #633, opened 2024-07-19
Fix missing base in redirect target

### LINKED ISSUE #632: Redirect does not work with a custom base
If you configure a base like `my-company` the redirect from `/` to `/docs` gonna look like `<config.homepageLink>/docs` instead of `<config.homepageLink>/my-company/docs`.

I think this is because the static redirects in astros config do not consider the base for the target, see https://github.com/withastro/astro/issues/7774

otbe:
closes #632


boyney123:
Thanks again @otbe LGTM

PR #656, opened 2024-08-01
fix/missing-base-url-at-assets

### LINKED ISSUE #637: OpenAPI file not found when using base
Hi,

as soon as we use the `<OpenApi/>` component together with `base`, the copied openapi.yml file is no longer found because the url is missing the base.

When visiting the Spec page it looks for `/generated/services/my-service/openapi.yml` but it should load it from `/<base>/generated/services/my-service/openapi.yml`

carlosallexandre:
## Motivation

It resolves #637 using `buildUrl` utility to generate the uri for public assets (openapi.yml, icons, favicon.ico).

yes.

### DIFF
--- a/.changeset/sixty-ants-jog.md
+++ b/.changeset/sixty-ants-jog.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+fix(core): fixed url paths for EC assets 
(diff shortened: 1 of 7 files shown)


boyney123:
Great, thanks for this!

PR #657, opened 2024-08-02
fix: remove forward slash before # to prevent double trailing slash

### LINKED ISSUE #638: With the trailing Slash Fragment identifiers no longer work
As a Follow on from [#569](https://github.com/event-catalog/eventcatalog/issues/569) when you enable the trailing slash as documented [here](https://www.eventcatalog.dev/docs/api/config#trailingSlash) any fragment identifiers are now broken as they will add a double / before the fragment
image

carlosallexandre:
## Motivation

It removes the forward slash at the anchor element on `DocsNavigationSidebar` preventing double trailing slash when trailingSlash is enabled.

Resolves #638.

Yes.

### DIFF
--- a/.changeset/tender-maps-walk.md
+++ b/.changeset/tender-maps-walk.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+core(fix): removed forward slash before # to prevent double trailing slash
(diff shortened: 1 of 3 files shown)


boyney123:
Thank you!

PR #685, opened 2024-08-06
chore(core): setup vitest and fixing tests.

boyney123:


boyney123:
Shout out to @carlosallexandre for helping here!

PR #688, opened 2024-08-08
fix: cleanUrl must only replace double slash in URL pathname

c-wygoda:
## Motivation

URLs in anchors in HTML are invalid as cleanUrl also replaces the double slash in the schema part instead of only in the pathname, which causes links to be broken.

No, because that file does not exist.

### DIFF
--- a/src/utils/url-builder.ts
+++ b/src/utils/url-builder.ts
@@ -1,7 +1,9 @@
 import config from '@eventcatalog';
 
-const cleanUrl = (url: string) => {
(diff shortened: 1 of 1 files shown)


boyney123:
Thanks @c-wygoda

boyney123:
@all-contributors add @c-wygoda for code

boyney123:
We have a failing build on this at the moment, any chance you look @c-wygoda ?

PR #698, opened 2024-08-09
Fix 697 : Truncate labels on visualiser page

debae:
Added truncate as class to BasicList to also visualise long service/event/domain names correctly on visualiser page.


boyney123:
Thank you @debae !

PR #704, opened 2024-08-09
feat: add semantic version range

### LINKED ISSUE #646: Allow a service to declare it can receive multiple versions of a message
Currently we need to be very specific when we declare the version of a message a service can receive. 

I find this to be somewhat a limit, because when a producer changes something in the events it emits, usually does it in a compatibile way and all the consumers can continue to work without any issue.

Producers can - and usually do - respect [semver](https://semver.org/), signaling compatibility breaks by increasing the major.

I think it would be nice to allow to specify versions a service can handle with something similar to what npm does with package dependencies (see [here](https://github.com/npm/node-semver#versions)), or event a subset of that.

carlosallexandre:
It adds semantic version range to specify receives/sends with valid version range.

## Motivation

Resolves #646

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/event-catalog/eventcatalog/blob/main/CONTRIBUTING.md#pull-requests)

### DIFF
--- a/.changeset/shiny-dodos-develop.md
+++ b/.changeset/shiny-dodos-develop.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": minor
+---
+
+feat(c0re): add semantic version support, referencing services or messages can now be done with semver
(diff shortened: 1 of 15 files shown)


boyney123:
Nice! Give me some to time to review/test this, but I love the idea.

PR #715, opened 2024-08-12
feat(core): added new example and fixes various bugs

boyney123:
Includes

- Adding new example for insurance domain
- Fix for flows default data when no flows are added (bug)
- Watcher and astro copy logic fixed


scripts/watcher.js
```diff
@@ -12,57 +12,74 @@ const catalogDirectory = process.env.CATALOG_DIR;
...
-      const { path: eventPath, type } = event;
-      const file = eventPath.split(item)[1];
-      let newPath = path.join(contentPath, item, extensionReplacer(item, file));
+      const { path: changedFile, type } = event;
+
+      // normalize path for OS (windows/mac)
+      const normalizedPath = path.normalize(changedFile);
+
+      // Split parts
+      const parts = normalizedPath.split('/').reverse();
```
carlosallexandre, line 52: use `path.sep` rather than `'/'`

PR #718, opened 2024-08-12
Feature/615 add ability to see how many domains a team and users own

debae:
- added owned domains to both users and teams
- fixed wrong title for owners of domain


boyney123:
Thanks @debae !

PR #722, opened 2024-08-13
Fix #700 : Only show latest versions on visualiser page

### LINKED ISSUE #700: Only show latest versions in visualizer sidebar
At the moment visualizer shows all the versions of all resources. I think we should only show the latest version of a resource.

debae:
Change in sidebar of visualiser to only show latest versions


boyney123:
Thanks @debae , going to add to the next minor version of EC I have building up with the support for flows. Hopefully aiming to release at end of week.

PR #728, opened 2024-08-16
fix: bind messages to producers/consumers when version is specified as latest/undefined

carlosallexandre:
Includes:
- It fixes the bind between events/commands when the services (producers/consumers) specifies the version of sends/receives as latest/undefined.
- Support for other types of version, such as v1 || 1 rather than only semantic version. [Requested here](https://discord.com/channels/918092420338569216/929038812175278091/1273375102557491270) by @otbe


boyney123:
Thanks again @carlosallexandre , sorry taken me awhile to review this, thanks for adding all the tests too. Very much appreciated

PR #733, opened 2024-08-25
Change home redirect status code from permanent to temporary

dreglad:
This PR suggests making the home redirect temporary (status 302) rather than permanent (301, used by default).

#### Rationale
The default permanent redirect makes it hard to change the `landingPage` config setting since browsers tend to cache 301 redirects, well.. permanently. Any future changes may be ignored by the browser,  leading to unexpected behavior.


boyney123:
Thank you @dreglad makes sense to me!

PR #736, opened 2024-08-26
fix: look for slug instead of id

### LINKED ISSUE #668: Changelogs are not detected if id of message is not identical to folder name
I was eager implementing the recently introduced changelogs in my catalog however I noticed that all my events just show `No changelogs found.`.

I tracked it down to the fact that [here](https://github.com/event-catalog/eventcatalog/blob/main/src/utils/changelogs/changelogs.ts#L11) `log.id` is basically the file path to the `changelog.mdx` but `data.id` is the id.

```
log.id: events/Inventory/InventoryAdjusted/versioned/1.0.0/changelog.mdx
collection: events
data.id: inventoryadjusted
```

otbe:
## Motivation

closes #668 

I tested it for "inventoryadjusted" (thats why I changed the ID there) however versioning in general seems to be a bit broken.
It works for domains but changelog page is empt and it does not work at all for services (also no changelogs). Not quite sure since when and at least this is what happens with the default example. Can you verify? (I always have issues running this)

Yes

### DIFF
--- a/examples/default/events/Inventory/InventoryAdjusted/index.md
+++ b/examples/default/events/Inventory/InventoryAdjusted/index.md
@@ -1,5 +1,5 @@
 ---
-id: InventoryAdjusted
+id: inventuryadjusted
 name: Inventory adjusted
 version: 1.0.1
 summary: |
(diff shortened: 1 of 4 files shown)


boyney123:
Thank you @otbe , not ignoring this! Will take a look later half of this week, thanks mate!

otbe:
Anything I can do here? :D @boyney123 
Last missing bit to finally switch to ECv2 for us 🥳

boyney123:
Thanks @otbe sorry for the turn around. I fixed in #772, same fix as yours without the id changes.

Not ideal long term, need a better solution, but happy to merge this to get you unblocked!

PR #737, opened 2024-08-29
fix: invalid hook call warning

### LINKED ISSUE #726: Fix "Invalid hook call" in logs for EventCatalog
image

When running EventCatalog I see this in the logs. Be great to debug this and fix if we can.

carlosallexandre:
It replaces the client:load by client:only directive of NodeGraph and SchemaViewer to prevent these from render at build time on the server.

Fix #726


boyney123:
Thank you for picking this up, LGTM 🙏

PR #749, opened 2024-09-01
fix(core): splitting target path while coping content files

devpsk:
Hi,

I noticed a bug when executing the command: `npm run build` aka (`eventcatalog build`). When I try to build the production version of the catalog, one of my application/microservice disappears. The application that disappears in the production environment _(everything works on localhost)_ is the one named **external-services-builder**

During analysis, I found that the method `getTargetPath` is losing my name. The directory where Event Catalog places applications is **services** and my application / microservice also contains the word "services" _(external-services-builder)_. So after calling the method "getTargetPath," an incorrect path is generated, which causes the application not to appear in the menu at all.

I analyzed the change history in the file `scripts/catalog-to-astro-content-directory.js` and noticed that it used to work previously. I restored that one code fragment.

### DIFF
--- a/scripts/catalog-to-astro-content-directory.js
+++ b/scripts/catalog-to-astro-content-directory.js
@@ -7,10 +7,17 @@ import os from 'node:os';
 const __filename = fileURLToPath(import.meta.url);
 const scriptsDir = path.dirname(__filename);
 
+const cleanRelativePath = (relativePath, type) => {
+  const position = relativePath.indexOf(`${type}${path.sep}`);
+
+  return relativePath.slice(position + `${type}${path.sep}`.length);
+};
+
 const getTargetPath = (source, target, type, file) => {
   const relativePath = path.relative(source, file);
-  const cleanedRelativePath = relativePath.split(`${type}/`);
-  const targetForEvents = path.join(type, cleanedRelativePath[1]);
+  const cleanedRelativePath = cleanRelativePath(relativePath, type);
+  const targetForEvents = path.join(type, cleanedRelativePath);
+
   return path.join(target, targetForEvents);
 };
 
@@ -64,9 +71,9 @@ const copyFiles = async ({ source, target, catalogFilesDir, pathToMarkdownFiles,
     }
 
     const relativePath = path.relative(source, file);
-    const cleanedRelativePath = relativePath.split(`${type}/`);
-    if (!cleanedRelativePath[1]) continue;
-    const targetForEvents = path.join(type, cleanedRelativePath[1]);
+    const cleanedRelativePath = cleanRelativePath(relativePath, type);
+    if (!cleanedRelativePath) continue;
+    const targetForEvents = path.join(type, cleanedRelativePath);
 
     // Catalog-files-directory
     const targetPath = path.join(catalogFilesDir, targetForEvents);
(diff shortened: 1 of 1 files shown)


scripts/catalog-to-astro-content-directory.js
```diff
@@ -9,7 +9,7 @@ const scriptsDir = path.dirname(__filename);
 
 const getTargetPath = (source, target, type, file) => {
   const relativePath = path.relative(source, file);
-  const cleanedRelativePath = relativePath.split(type);
+  const cleanedRelativePath = relativePath.split(`${type}/`);
```
carlosallexandre, line 12: The `/` was removed for compatibility with `win32` plataform.
Maybe `path.sep` could works 🤷🏼‍♂️. 

When I worked at a refactor proposal for the `catalog-to-astro-content-directory.js` I got the following for this snippet:
```ts
/**
 * Check if the key is an ASTRO COLLECTION KEY
 * @param {string} key
 * @returns {boolean}
 */
const isCollectionKey = (key) => {
  const COLLECTION_KEYS = ['commands', 'domains', 'events', 'pages', 'services', 'teams', 'users'];
  /** 
   *  Using includes won't works for your use case either
   *  The solution will be a comparison
   */
  return COLLECTION_KEYS.includes(key);
};

/**
 * Generates the path until the ASTRO_COLLECTION_KEY or the PROJECT_DIR root.
 * @param {string} filePath The path to the file.
 * @returns {string} The path until the COLLECTION_KEY or PROJECT_DIR root.
 */
const getRelativeTargetPath = (filePath) => {
  const filePathParsed = path.parse(filePath);
  const fileDir = filePathParsed.dir.split(path.sep).filter(Boolean);

  const relativePath = [];
  for (let i = fileDir.length - 1; i >= 0; i--) {
    relativePath.unshift(fileDir[i]);
    if (isCollectionKey(fileDir[i])) break;
  }

  return path.join(...relativePath, filePathParsed.base);
};
```
You can check the implementation [here](https://gist.github.com/carlosallexandre/9a410d493b4f4f63d54157642395381d). 
I hope this can help you find a suitable solution.

devpsk: Thanks @carlosallexandre for review. I committed new changes, can you check?

scripts/catalog-to-astro-content-directory.js
```diff
@@ -7,10 +7,17 @@ import os from 'node:os';
 const __filename = fileURLToPath(import.meta.url);
 const scriptsDir = path.dirname(__filename);
 
+const cleanRelativePath = (relativePath, type) => {
+  const position = relativePath.indexOf(`${type}${path.sep}`);
```
carlosallexandre, line 11: Take the following path `/root/something-events/other/events/my-event/index.md` and type equals `events`. The `indexOf` will match the first `events/` which is not what we expect. 

As pointed [here](https://github.com/event-catalog/eventcatalog/pull/812#issuecomment-2356743840) add `path.sep` before the type too.

carlosallexandre (changes requested):

scripts/catalog-to-astro-content-directory.js
devpsk: I added. Please review again.

devpsk: Bump, @carlosallexandre

boyney123:
Thank you @devpsk 

This should be fixed in 2.9.1, give that ago, let us know if you still have issues!

Thank you for the PR btw!

PR #762, opened 2024-09-09
refine and relook rapi-doc openapi renderer

XaaXaaX:
## Motivation
having OpenApi spec viewer with a higher level of look and feel to make it readable but operational.

- [x] Enable Try button
- [x] Enable Auth configuration
- [x] Refine them and styling
- [x] Expand the schema tab by default instead of example
- [x] Set schema representation as tabular

### DIFF
--- a/.changeset/six-dolphins-protect.md
+++ b/.changeset/six-dolphins-protect.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+chore(core): fixed styled for openapi pages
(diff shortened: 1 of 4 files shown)


boyney123:
npm run `npm run format`

boyney123:
Thank you mate, I'm tempted to get this in at the moment, cos it fixed what we have, but also review the other UI too along side it :)

boyney123:
One thing I don't like about spotlight at the moment is this too 
image

glc-omid:
> One thing I don't like about spotlight at the moment is this too 
> image
> 

If you look again here I find redoc very readable personally 
https://redocly.com/redoc

PR #772, opened 2024-09-11
fix(core): fixed changelogs path matching

boyney123:
Fix for #736


boyney123:
Think this is OK for now, thanks @otbe for the pull request, but I need to figure out a proper way to do this without relying on the paths like this in the future. Happy to merge to unblock you!

PR #775, opened 2024-09-12
dont generate discover pages for each item in a type collection

### LINKED ISSUE #774: Massive performance issues after upgrading to v2.7.1
After upgrading to v2.7.1 our build basically does not finish anymore (in a reasonable time :D).

I think I tracked it down to the changes introduced in [v2.6.3](https://github.com/event-catalog/eventcatalog/commit/d35846ba6d3e5d5ff12c9ac0304f5c3940e08ee5). 
I can see the output 

```
09:02:32   ├─ /discover/events/index.html (+3.36s)
```
like many 100s of times. If you compare the code of [src/pages/disover/[type]/index.astro](https://github.com/event-catalog/eventcatalog/blob/main/src/pages/discover/%5Btype%5D/index.astro#L9) to the version [before](https://github.com/event-catalog/eventcatalog/blob/067fd89af6bc0a71cc7a0686f250ac8bb563878d/src/pages/discover/%5Btype%5D/index.astro) and after the change introduced in v2.6.3 it seems like we did the page generation for each type and and now for each element of a type (so basically for each event instead of for all of them).

otbe:
## Motivation

closes #774 

yes

### DIFF
--- a/.changeset/kind-geese-type.md
+++ b/.changeset/kind-geese-type.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+fix(core): only generate discover page for each collection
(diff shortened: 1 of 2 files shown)


boyney123:
O wow! Thanks for this 😅

PR #812, opened 2024-09-17
fix(script/watcher): support being in a subdirectory named events

gabrielcolson:
## Motivation

When the path of a catalog project contains `events`, `services`, `commands`, etc., the watcher crashes on file change.


carlosallexandre:
This is something that must be addressed [here](https://github.com/event-catalog/eventcatalog/blob/main/scripts/catalog-to-astro-content-directory.js#L13) too. As pointed [here](https://github.com/event-catalog/eventcatalog/pull/749)

boyney123:
I believe this is now fixed, feel free to open again if it is not.

PR #820, opened 2024-09-17
feat/full-diagram

### LINKED ISSUE #760: Update "Download Visual" to download full diagram
> FYI - It doesn't look you have any issue templates to select between. https://github.com/event-catalog/eventcatalog/issues/new/choose links to creating a new issue, not selecting a template. You need to create an ISSUE_TEMPLATE folder within .github that contains the different issue templates. This will then prompt users to select a template when creating an issue

Currently when you try to download visualisations e.g. a flow, it downloads an image that only contains the nodes that are currently visible on the page. When creating large it means that much of the diagram is cut off.

It would be great if the "Download Visual" button instead downloads the entire diagram.

carlosallexandre:
## Motivation

Closes #760

### DIFF
--- a/.changeset/rare-papayas-lay.md
+++ b/.changeset/rare-papayas-lay.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+fix(core): downloading images in visualizer now shows the whole diagram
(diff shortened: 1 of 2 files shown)


boyney123:
Thank you @carlosallexandre , just tested and looks good to me! Thanks for picking this up 🙏

PR #821, opened 2024-09-18
add path an verb in openapi viewer nav bar

XaaXaaX:
## Motivation
In openApi viewer, the summary was shown in nav bar, and as this became verbose  and hard to read , the option of showing the http verb + endpoint path in navbar seems a better and readable option


boyney123:
Just had a look, looks great to me! Thanks @XaaXaaX happy to merge this

PR #833, opened 2024-09-29
chore/eventcatalog-config-import

### LINKED ISSUE #830: Figure out a better way to deal with CJS and ESM for eventcatalog.config.js
### Use Case

Quite a hard problem to explain but will do my best.

Most of EventCatalog us built using TypeScript and compiles into CJS files (bin folder) etc. This uses tsup.

When people checkout and use EventCatalog the scripts are all compiled into CJS files. Means when we run build, dev etc, the files are commonJS. Not a problem....

But these scripts import the EventCatalog file to understand users settings. The EventCatalog file itself is written using ESM..... example here.

```
/** @type {import('../../bin/eventcatalog.config').Config} */
export default {
  cId: '8027010c-f3d6-417a-8234-e2f46087fc56',
  title: 'EventCatalog',
  tagline: 'Discover, Explore and Document your Event Driven Architectures',
  organizationName: 'Awesome Fake Company',
  homepageLink: 'https://eventcatalog.dev/',
  editUrl: 'https://github.com/event-catalog/eventcatalog/edit/main',
  port: 3000,
  logo: {
    alt: 'EventCatalog',
    src: '/logo.png',
    text: 'EventCatalog',
  },
  base: '/',
  trailingSlash: false,
  docs: {
    sidebar: {
      // Should the sub heading be rendered in the docs sidebar?
      showPageHeadings: true,
      services: {
        visible: true,
      },
      messages: {
        visible: true,
      },
      domains: {
        visible: true,
      },
      teams: {
        visible: true,
      },
      users: {
        visible: true,
      },
    },
  }
};


```

As you can see we are using ESM exports, and also folks can use packages and imports here from ESM syntax. Like our generators....

```
import path from 'path';
import url from 'url';

const __dirname = path.dirname(url.fileURLToPath(import.meta.url));

/** @type {import('@eventcatalog/core/bin/eventcatalog.config').Config} */
export default {
  title: 'OurLogix',
  tagline: 'A comprehensive logistics and shipping management company',
  organizationName: 'OurLogix',
  homepageLink: 'https://eventcatalog.dev/',
  landingPage: '',
  editUrl: 'https://github.com/boyney123/eventcatalog-demo/edit/master',
  // By default set to false, add true to get urls ending in /
  trailingSlash: false,
  // Change to make the base url of the site different, by default https://{website}.com/docs,
  // changing to /company would be https://{website}.com/company/docs,
  base: '/',
  // Customize the logo, add your logo to public/ folder
  logo: {
    alt: 'EventCatalog Logo',
    src: '/logo.png',
    text: 'OurLogix',
  },
  docs: {
    sidebar: {
      // Should the sub heading be rendered in the docs sidebar?
      showPageHeadings: true,
    },
  },
  generators: [
    // Add single AsyncAPI file to a domain
    [
      '@eventcatalog/generator-asyncapi',
      {
        services: [
          { path: path.join(__dirname, 'asyncapi-files', 'orders-service.yml')}
        ],
        domain: { id: 'orders', name: 'Orders', version: '0.0.1' },

        // Run in debug mode, for extra output, if your AsyncAPI fails to parse, it will tell you why
        debug: true
      },
    ],
    // Add many AsyncAPI files to a domain
    [
      '@eventcatalog/generator-asyncapi',
      {
        services: [
          { path: path.join(__dirname, 'asyncapi-files', 'payment-service.yml')}
          { path: path.join(__dirname, 'asyncapi-files', 'fraud-detection-service.yml')}
        ],
        domain: { id: 'payment', name: 'Payment', version: '0.0.1' },
        // Run in debug mode, for extra output, if your AsyncAPI fails to parse, it will tell you why
        debug: true
      },
    ],
  ],
};
```

### The problem

When the compiled bin/scripts try and get the EventCatalog.config.js file, it fails as it does not understand this syntax.So we have a script to fix this (https://github.com/event-catalog/eventcatalog/blob/main/scripts/eventcatalog-config-file-utils.js), which is abit of a hack.... this transforms the catalog file back to `.cjs` so we can read the contents... and removes it.

One fix would be forcing eventcatalog.config.js file to be CJS syntax.... but Astro also reads this file..... and expects ESM..... 

### Proposed Solution

Not really sure, but figure out a way to we can import the Eventcatalog.config.js file without the hacks found in here https://github.com/event-catalog/eventcatalog/blob/main/scripts/eventcatalog-config-file-utils.js

I want to give users the ability to use any external package in there config file and have catalog scripts and project be able to read there file....

Any ideas would be welcome!

### Implementation Notes

_No response_

### Community Notes

carlosallexandre:
## Motivation

Close #830.

## Solution

It imports the `package.json` from projectDirectory to check the module system, ECMAScript Modules or CommonJS. If the  module system is ESM then just import the `eventcatalog.config.js`. Otherwise, make a copy of eventcatalog.config.js to the os tmpdir/eventcatalog.config.mjs and import it. 

It removes the convertEsmToCjs used before.

This solution was chosen because it touches less code.

### DIFF
--- a/.changeset/good-rockets-study.md
+++ b/.changeset/good-rockets-study.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+chore(core): changed the way eventcatalog file is now imported into the project
--- a/scripts/__tests__/eventcatalog-config-file-utils.spec.ts
+++ b/scripts/__tests__/eventcatalog-config-file-utils.spec.ts
@@ -10,6 +10,7 @@ import {
   verifyRequiredFieldsAreInCatalogConfigFile,
   writeEventCatalogConfigFile,
 } from 'scripts/eventcatalog-config-file-utils';
+import { tmpdir } from 'os';
 
 describe('catalog-to-astro-content-directory', () => {
   afterEach(async () => {
@@ -101,11 +102,11 @@ describe('catalog-to-astro-content-directory', () => {
         })
       );
     });
-    it('eventcatalog.config.cjs (temp commonjs file) is removed when fetching the catalog file', async () => {
+    it('removes eventcatalog config from the tmpdir', async () => {
       await getEventCatalogConfigFile(CATALOG_DIR);
 
-      const commonJsFile = path.join(CATALOG_DIR, 'eventcatalog.config.cjs');
-      expect(existsSync(commonJsFile)).toBe(false);
+      const configFile = path.join(tmpdir(), 'eventcatalog.config.mjs');
+      expect(existsSync(configFile)).toBe(false);
     });
   });
 
--- a/scripts/__tests__/example-catalog/package.json
+++ b/scripts/__tests__/example-catalog/package.json
@@ -1,0 +1,3 @@
+{
+  "description": "Only for testing purposes"
+}
--- a/scripts/eventcatalog-config-file-utils.js
+++ b/scripts/eventcatalog-config-file-utils.js
@@ -1,38 +1,13 @@
 import { readFile, writeFile, rm } from 'node:fs/promises';
-import { existsSync } from 'fs';
+import { existsSync } from 'node:fs';
+import { copyFile } from 'node:fs/promises';
 import path from 'node:path';
 import { v4 as uuidV4 } from 'uuid';
 import { pathToFileURL } from 'url';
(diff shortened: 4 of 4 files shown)


boyney123:
Thanks @carlosallexandre , much cleaner!

Issue #834, opened 2024-09-30
Events containing "events" in their name fails to be shown in event catalog


HansMartinJ:
### I tried this:

created event directory called "foo-bar-events/"

### This happened:

it wasn't shown in event catalog

### I expected this:

for it to be shown

### Is there a workaround?

not have the event contain "events"

### Anything else?

The problem is in `getTargetPath` in `core/scripts/catalog-to-astro-content-directory.js`. More specifically with the split

### EventCatalog Version

2.8.1

### Node.js Version

v22.8.0

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

devpsk:
I have a similar situation, look at: https://github.com/event-catalog/eventcatalog/pull/749/files

boyney123:
Hey, thanks for raising this. This should be fixed in 2.9.1, give that ago, and feel free to reopen the issue if you still have problems, with instructions how we can replicate it. Thanks!

PR #835, opened 2024-09-30
fix(core): fix events in event name bug

HansMartinJ:
Fixes issue #834


HansMartinJ:
Closing it because I see you are already working on a solution and this didn't seem to work

PR #845, opened 2024-10-04
Fix/semver build type missing error

XaaXaaX:
## Motivation
The catalaog build raise an error indicating the semver types are missing 

Screenshot 2024-10-04 at 23 05 02


carlosallexandre:
`@types/semver` should be a direct dependency of the `eventcatalog-core` to avoid this error.

It's strange that `Verify install with dependencies` github action catches this error only now.

XaaXaaX:
> `@types/semver` should be a direct dependency of the `eventcatalog-core` to avoid this error.
> 
> It's strange that `Verify install with dependencies` github action catches this error only now.

@carlosallexandre oh yes you are right, for the direct dependency. But I was also curious to see why it fails now,

XaaXaaX:
it seems working now, just curious where the semver is used in eventcatalog 
i see that changesets has a dependency on semver and @types/semver but also astro has a semver dependency but not @types/semver

to me the changesets cli must be a dev dependency in eventcatalog and not a dependency 
this PR solves the problem but I would like to know your ideas @carlosallexandre @boyney123 any idea?

carlosallexandre:
@XaaXaaX 

`semver` was added to support #646.

I agree with `@changeset/cli` must be a dev dependency.

XaaXaaX:
> @XaaXaaX
> 
> `semver` was added to support #646.
> 
> I agree with `@changeset/cli` must be a dev dependency.

Done

boyney123:
thanks folks, this is now in latest version, got added as another PR somewhere. Thanks for raising and fixing.

PR #846, opened 2024-10-07
fix: Update dynamic data in index.astro

d-o-h:
Update the dynamic data in the index.astro file to display the correct counts for domains, services, commands, events, and flows. The data is now fetched dynamically from the respective arrays (domains, services, commands, events, and flows) instead of using hardcoded values.


boyney123:
@all-contributors add @d-o-h for code

PR #850, opened 2024-10-09
Fix/react peer conflicts

XaaXaaX:
## Motivation

EC uses many dependent packages and we see lots of verbose warnings in CI 
We can override the react peers to use the project react installed version 
Screenshot 2024-10-09 at 13 49 34

## Caution
This need to be tested and we need to validate if all parts of project are working

### DIFF
--- a/.changeset/chilled-bugs-pay.md
+++ b/.changeset/chilled-bugs-pay.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+chore(core): removing react conflicts with installs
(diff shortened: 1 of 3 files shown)


XaaXaaX:
> Yeah all these react peer conflicts from spotlight are slightly annoying. I just tested locally and things to work with this, but I'm not sure if we have any side effects introducing this? Any thoughts @carlosallexandre ?

I did also a walkthrough and sound promising but you know , nothing clear in JS ecosystem :D

carlosallexandre:
LGTM,

But if you are afraid, we can place some e2e tests. I setup the playwright for e2e tests right now - #873.

boyney123:
I'm going to close and reopen this when the time comes, just tidying up on some old PR/issues

PR #853, opened 2024-10-10
Feat/fix domain visualization

XaaXaaX:
## Motivation
The domain visualizer rendering and layout has some overlapping and incostistent positioning, The domain viz is the most complext one in EC and it shows a more global landscape than service or messages do 

The solution was to use the same calculate Node in @utils that consider defining the position of each node in graph 

### Old results
Screenshot 2024-10-10 at 11 32 24

### New Results
Screenshot 2024-10-10 at 11 29 47


boyney123:
Very nice! Much better <3

boyney123:
Thank you @XaaXaaX

PR #856, opened 2024-10-10
Feat/refine docs sidebar

XaaXaaX:
## Motivation
The nav sidebar becomes complicated to navigate , as a first step adding the possibility to collapse groups can be handy
Screenshot 2024-10-10 at 17 18 49


boyney123:
Thanks for this! Going to close it as we have the new sidebar in the latest version. Thanks for the idea, I moved it into the latest version of the UI.

PR #859, opened 2024-10-11
chore/url-builder-client-server-compatible

### LINKED ISSUE #847: Unable to search in discover tab, also unable to click next page
### I tried this:

Generated the catalog using generator-asyncapi (5 events,  8 services). 

Went into the discover page and clicked on the search fields of events. 

### This happened:

It successfully shows a preview of all the available events, but when I click on one of them to filter, nothing happens.

### I expected this:

For it to filter based on the even I chose

### Is there a workaround?

_No response_

### Anything else?

This also applies for filtering services and domains as well

Also unable to click next page (even when I had more than 8 services)

generator-asyncapi version: 2.1.1

### EventCatalog Version

2.8.1

### Node.js Version

v22.8.0

### Platform(s)

_No response_

### Community Notes

carlosallexandre:
## Motivation

After investigating the issue #847, I found that `url-builder` is imported by a client component (`Table` component used by `discover page`). As `url-builder` depends on the `trailingSlash` config to generate the appropriately url, running the `url-builder` in  browser environments causes the error mentioned in the issue.

A few client components use `url-builder-client` that doesn't rely on the `@eventcatalog.config` but needs the `trailingSlash` as a param. This brings a burden to the client component as it needs to know the `trailingSlash` needed only to the `url-builder` that is passed by a server component through props. 

So, to remove the unneeded trailingSlash props and avoiding the duplication of `url-builder` this PR adds a global variable (`__EC_TRAILING_SLASH__`)  that can be acessed by the `url-builder` in node/browser environments without any issue.

Fix #847.


boyney123:
Thank you @carlosallexandre such a better implementation and idea, didnt even know this was possible. 

Thank you mate!

Issue #849, opened 2024-10-08
Config setting for configuring the Astro outDir


Sheemap:
### Use Case

We use EventCatalog in a monorepo, where everything is packaged and stuck into a `dist/{project_path}` dir. This involves running a process and then copying over specific out files/dirs.

Unfortunately, because we cannot customize the path of EventCatalog's dist, when we run our packaging tool, it ends up in `dist/dist`

### Proposed Solution

It would be awesome if we could just specify a custom `outDir` in eventcatalog.config.js. Similar to the other astro settings available, like `trailingSlash`

### Implementation Notes

Id imagine adding the `outDir` follows the same patterns as other astro settings. Accepting a value from eventcatalog.config.js, while having a default for `dist`

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out in the [EventCatalog Discord](https://discord.com/invite/3rjaZMmrAm).

boyney123:
Hey @Sheemap 

Added `outDir` to the catalog in version `@eventcatalog/core@2.11.4`.

https://www.eventcatalog.dev/docs/api/config#outDir

Thanks for raising the issue!

boyney123:
Hey @Sheemap 

Thanks for coming back, ahh yes makes sense. TBH I think this makes sense to me

`outDir: config.outDir ? join(projectDirectory, config.outDir) : join(projectDirectory, 'dist'),`

I can't see why the core would ever need it, and just write to the users catalog makes sense. I will raise a PR for this and clean up the scripts as you said :) thanks for feedback!

boyney123:
OK think I have a fix here #915 will test it out

PR #860, opened 2024-10-13
feat: add outDir config setting

Sheemap:
## Motivation

This is to address https://github.com/event-catalog/eventcatalog/issues/849, allowing users the ability to configure which directory astro outputs build files.

Let me know what yall think! :) Thank you!

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -10,7 +10,7 @@ import { join } from 'node:path';
 import config from './eventcatalog.config';
 import expressiveCode from 'astro-expressive-code';
 
-const coreDirectory = process.env.CATALOG_DIR || process.cwd();
+const projectDirectory = process.env.PROJECT_DIR || process.cwd();
 const base = config.base || '/';
 
 // https://astro.build/config
@@ -18,7 +18,7 @@ export default defineConfig({
   base,
   server: { port: config.port || 3000 },
 
-  outDir: join(coreDirectory, 'dist'),
+  outDir: join(projectDirectory, config.outDir || 'dist'),
 
   // https://docs.astro.build/en/reference/configuration-reference/#site
   site: config.homepageLink || 'https://eventcatalog.dev/',
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -2,8 +2,8 @@ import { defineConfig } from 'astro/config';
 import tailwind from '@astrojs/tailwind';
 import mdx from '@astrojs/mdx';
 import react from '@astrojs/react';
-import pagefind from "astro-pagefind";
-import { mermaid } from "./src/remark-plugins/mermaid"
+import pagefind from 'astro-pagefind';
+import { mermaid } from './src/remark-plugins/mermaid';
 import { join } from 'node:path';
 
 /** @type {import('bin/eventcatalog.config').Config} */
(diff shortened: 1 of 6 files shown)


astro.config.mjs
```diff
@@ -18,13 +18,13 @@ export default defineConfig({
   base,
   server: { port: config.port || 3000 },
 
-  outDir: join(coreDirectory, 'dist'),
+  outDir: join(coreDirectory, config.outDir || 'dist'),
```
carlosallexandre, line 21: This `outDir` refers to `.eventcatalog-core`. Is this `dir` you want to change or [this](https://github.com/event-catalog/eventcatalog/blob/main/bin/eventcatalog.ts#L114)?

Sheemap: Im not sure if that works. If I revert my changes  and then change the line you linked to `copyFolder(join(core, 'dist'), join(dir, 'dist-wahoo'));`, then run `npm run build`, and it still outputs it to `dist`. 

But if I change it in this astro config, it does actually build it into the correct folder

carlosallexandre: To test this you need to do something similar to [this docker image](https://github.com/event-catalog/eventcatalog/blob/main/Dockerfile), that have been placed to catch bugs of this package on production (user's directory).

What I do when I want to test here:
1. run `npm run build:bin` - _builds the `eventcatalog`_
2. run `npm pack` - _generates the package that will be publish at npm registry_
3. `cd examples/default/` - _goes to the example emulating the user's directory_
4. `npm install ../../*.tgz` - _install the package generated on step 2_
5. `npx eventcatalog build` or `npx . build` - _runs the build command_

With this steps we can check if the `dist-wahoo` we be in `examples/default/`.

Sheemap: Oh I see! Thank you for taking the time to walk me through that. I will test and make some changes later tonight

Sheemap: Okay I got it working, thanks again. I definitely understand this codebase more after looking at those.

Though now Im getting a bit blocked on how to actually access that `outDir` value from within the npx build action. Do you have any guidance there?

If we arent able to access the config (or its a pain for some reason), perhaps we could modify the astro config still, and output to something like `astro-dist/${config.outDir}`

If we clear `astro-dist` before each build, then the single folder in there should be our target. Which we can just copy two levels up, no renaming necessary.

Fairly uninformed thoughts here, but hey :grin:  What do you think?

carlosallexandre: > Though now Im getting a bit blocked on how to actually access that `outDir` value from within the npx build action. Do you have any guidance there?

You can access the `eventcatalog.config` using [this](https://github.com/event-catalog/eventcatalog/blob/main/scripts/eventcatalog-config-file-utils.js#L15)



> If we clear `astro-dist` before each build, then the single folder in there should be our target. Which we can just copy two levels up, no renaming necessary.

This is a fair point.

Another random thought about it. If we do something like below:
image

We generate the build direct at the user's directory avoiding [this copy](https://github.com/event-catalog/eventcatalog/blob/main/bin/eventcatalog.ts#L114). What could be wrong with that???

CATALOG_DIR usually is `.eventcatalog-core`
PROJECT_DIR is the user's directory - usually `process.cwd()`

Sheemap: Smart! I like that idea. I just tested it a bit locally and pushed it up.

Unrelated to this, but `npx eventcatalog build` fails if Im not in the project dir. It would be nice to be able to use catalog commands at any level of the project

carlosallexandre: > Smart! I like that idea. I just tested it a bit locally and pushed it up.

Remove [this copy from build command](https://github.com/event-catalog/eventcatalog/blob/main/bin/eventcatalog.ts#L114) wasn't needed?


> Unrelated to this, but `npx eventcatalog build` fails if Im not in the project dir. It would be nice to be able to use catalog commands at any level of the project

What error you got it?

boyney123:
Thanks! I merged a copy of this in #849, thanks for the idea and the pull request!

PR #873, opened 2024-10-16
chore/config-e2e-tests

### LINKED ISSUE #868: Need some basic e2e tests in CI/CD
### Use Case

The project is getting more and more pull requests, and features. I think it's time to add some smoke tests or layer of e2e tests to the project. I'm not sure if we should use Cypress or Puppeteer, but something that can run quickly and start testing some basic pages etc, would be great. Something we add too over time.

### Proposed Solution

Implement a new GitHub action for e2e tests.

### Implementation Notes

_No response_

### Community Notes

carlosallexandre:
## Motivation

Close #868

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -6,5 +6,6 @@ export default getViteConfig({
   plugins: [tsConfigPaths()],
   test: {
     globals: true,
+    include: ['**/__tests__/**/*.{test,spec}.?(c|m)[jt]s?(x)'],
   },
 });
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,11 +1,12 @@
 /// <reference types="vitest" />
 import { getViteConfig } from 'astro/config';
+import { defaultExclude } from 'vitest/config';
 import tsConfigPaths from 'vite-tsconfig-paths';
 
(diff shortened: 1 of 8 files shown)


vitest.config.ts
```diff
@@ -6,5 +6,6 @@ export default getViteConfig({
   plugins: [tsConfigPaths()],
   test: {
     globals: true,
+    include: ['**/__tests__/**/*.{test,spec}.?(c|m)[jt]s?(x)'],
```
boyney123, line 9: What does this do? Why is it required?

boyney123:
Awesome @carlosallexandre 👏 , great start to some e2e tests

vitest.config.ts
carlosallexandre: To prevent vitest run `e2e` tests. So, it will only look to tests inside dir `__tests__`.
We could use `exclude` too, if you prefer, excluding all test cases under `e2e/` and removing the include.

carlosallexandre: image
Using exclude

boyney123: Just moved it over 👍

PR #883, opened 2024-10-18
Chore/add big api in example

XaaXaaX:
## Motivation

This pull request add a big service based on a big api spec to help mesure the edge cases
The example found on the web


boyney123:
Wow sure is big thanks for this! I do wonder if this belongs in a separate repo tho?

XaaXaaX:
> Wow sure is big thanks for this! I do wonder if this belongs in a separate repo tho?

I just found it on redoc if I remember well 

May be have a new big one inspiring from this

XaaXaaX:
@boyney123 
i m on chat gpt to generate me an example enough big for a dummy service
It will be nice but not sure it will give enough big as this example we found ;)

XaaXaaX:
After playing with chatgpt it could not give me more than 500 lines with a limited set of endpoints ( it already fetched all my quota :D due to a long prompting try and fail :( ) 

@boyney123 
the example i used is from here https://github.com/Redocly/redoc/blob/main/demo/big-openapi.json

boyney123:
Closing in favor of #1036

PR #903, opened 2024-10-25
Fix/catalog home page counters

XaaXaaX:
## Motivation

Counters in EC Home Pages shows all available versions for each entity type, this PR fetch only the latest versions


boyney123:
Thank you @XaaXaaX ! Can you run `npm run format` on this PR? Looks like some format issues?

.github/dependabot.yml
```diff
@@ -0,0 +1,55 @@
+version: 2
```
boyney123, line 1: Guessing this file should not be part of this PR? Could you remove it? Thanks!

.changeset/early-steaks-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+feat(core): added api path and http verb in openapi viewer
```
boyney123, line 5:
```suggestion
fix(core): counters on homepage only include latest versions
```

boyney123:
Thank you @XaaXaaX looking good mate

PR #913, opened 2024-10-26
chore/remove-copy-files

carlosallexandre:
## Motivation

- [x] It removes copy eventcatalog config/styles to astro. These files are
copied by the hydration phase with the use o `mapCatalogToAstro`.
- [x] It removes unused `verify-build` script as the `build-ci` is used by the github actions (`npm run build:cd` refers to it)


boyney123:
Thanks @carlosallexandre , makes sense to me! Thanks for the PR!

PR #927, opened 2024-10-31
chore/cli-copy-core

carlosallexandre:
## Motivation

- [x] It avoid copying the core to the user's directory when the  `.eventcatalog-core` already exists and has the same version    as the `@eventcatalog/core`.
- [x] It adds a simple logger that wraps Console. Based on the logLevel enabled this logger decides between print or not the message.


bin/logger.ts
```diff
@@ -0,0 +1,45 @@
+type LoggerLevel = 'debug' | 'info' | 'warn' | 'error' | 'silent';
```
boyney123, line 1: Hey @carlosallexandre do you think it's worth using https://www.npmjs.com/package/pino for a logging lib? Pretty popular, saves us making our own?

boyney123: Good question, maybe add into this PR for now, unless there is lots to do, then another PR would be good then remove logging in this one, if that makes sense?

carlosallexandre: Sure...
I will work on this right now.

carlosallexandre:
Closing in favor #945

PR #935, opened 2024-11-06
feat(core): adding channels to eventcatalog

boyney123:
Adding ability to add optionally add channels between messages and services, includes a new resource type "channels".


examples/default/channels/inventory.events/index.md
```diff
@@ -0,0 +1,13 @@
+---
+id: inventory.events
+name: Inventory Events
+version: 1.0.1
+summary: |
+  The topic on which measured values may be produced and consumed.
+owners:
+    - dboyne
+address: smartylighting/streetlights/1/0/event/5/lighting/measured
+technology: kafka
```
XaaXaaX, line 10: Per Asyncapi spec this should be "protocol" as it mustly refers to servers object practically 

Per Cloudevents this is called also protocol but behind a bindings concept https://github.com/cloudevents/spec/tree/main/cloudevents/bindings

examples/default/channels/inventory.events/index.md
```diff
@@ -0,0 +1,13 @@
+---
+id: inventory.events
+name: Inventory Events
+version: 1.0.1
+summary: |
+  The topic on which measured values may be produced and consumed.
+owners:
+    - dboyne
+address: smartylighting/streetlights/1/0/event/5/lighting/measured
```
XaaXaaX, line 9: also i recommend look at specs and see if any thought naming can be found 

a channel often must have a host , but here as we talk abounts events inside a channel markdown maybe this can be rethinked

examples/default/channels/inventory.events/index.md
boyney123: Thanks! totally makes sense, funny I actually changed it the other day when I was working on it and just seen the comment. Thanks for the review/comment! Changed to protocol

examples/default/channels/inventory.events/index.md
boyney123: Hey @XaaXaaX thanks for the comment.

I'm not sure I 100% understand, any chance you could explain some more with maybe an example? Thanks mate!

XaaXaaX: Oh @boyney123 
Just i tried to say do we title the attribute address or host?
Just i m trying to adopt the same nomination provided by standard specifications 
But practically, i assume a channel can not be pointed only by a host

boyney123: Ahhh OK cool, yeah I took address from https://www.asyncapi.com/docs/reference/specification/v3.0.0#channelObject, and protocol from the same place, yeah anywhere we can stear towards standards I'm always down for!

PR #938, opened 2024-11-07
chore(core): fixing z-index issue with the header nav and removing the max-width

jacktomlinson:
Was taking a look at EC and running on my 27inch 4k Monitor - the catalog header starts to look at bit off as there is a max width set. This PR is now suggesting that we let this flow right to the edges (as the rest of the application does) so everything is in line. See screenshots below:

Before (https://demo.eventcatalog.dev):
image

After:
image

Also....

While I was here I noticed that there is a weird z-index issue where the payload content block goes above the header.

Before (https://demo.eventcatalog.dev/docs/events/OrderCancelled/0.0.1): 
image

After:
image

What you recon @boyney123 ?

### DIFF
--- a/.changeset/lovely-hairs-film.md
+++ b/.changeset/lovely-hairs-film.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+chore(core): fixing z-index issue with the header nav and removing the max-width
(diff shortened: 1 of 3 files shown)


boyney123:
Perfect thanks @jacktomlinson

PR #959, opened 2024-11-14
chore: move types deps to dev deps

carlosallexandre:
## Motivation

It move `@astrojs/check` to devDeps. Astro check is used at build-ci script only.
It move `@types/*` to devDeps as in 95ca9aa the types check was removed.


boyney123:
Thanks, makes sense to me!

boyney123:
Just tested it out, works great. Thanks for picking this up and finding it!

PR #960, opened 2024-11-15
chore/script to module

carlosallexandre:
## Motivation

It turns the following scripts into modules: `generate`, `log-build`, `watcher` and `hydrate`.

This way, we execute the modules inside the respectively cli commands instead of executing them with `node` defined into `package.json`. Doing that brings us the possibility to measure the time spent with each function and post to the analytics in future improvements.


boyney123:
Thank you @carlosallexandre , will take a look.

Any reason `default-files-for-collections` has been moved here? Maybe we can do that stuff in seperate PR (if required), or does this PR require those changes?

carlosallexandre:
> Any reason `default-files-for-collections` has been moved here? Maybe we can do that stuff in seperate PR (if required), or does this PR require those changes?

This makes the path to it into [`scripts/catalog-to-astro-content-directory.js`](https://github.com/carlosallexandre/eventcatalog/blob/chore/script-to-module/scripts/catalog-to-astro-content-directory.js#L81) more simple. 

Now, the tsup configuration build to `bin/dist/<files>` having as entrypoint the `bin/eventcatalog.ts`. As we not specify the `scripts/catalog-to-astro-content-directory.js` in the entrypoint it will be in the `bin/dist/eventcatalog.?(c)js`. So, from there the `default-files-for-collections` is 
```js
const __filename = fileURLToPath(import.meta.url);
const rootPkg = path.resolve(
  path.dirname(__filename),
  process.env.NODE_ENV === 'test' ? '../' : '../../'
);
const defaultCollectionFilesDir = path.join(rootPkg, 'default-files-for-collections');
``` 
otherwise, if we keep the `default-files-for-collections/` as is, it would be 
- structure
```
.
├── ...
├── bin                    
│   ├── dist
│   │   ├── eventcatalog.cjs
├── scripts                    
│   ├── default-files-for-collections
└── ...
```
- path in `scripts/catalog-to-astro-content-directory.js`
```js
const __filename = fileURLToPath(import.meta.url); // `bin/dist/eventcatalog.?(c)js`
const scriptsDir = path.resolve(
  path.dirname(__filename),
  process.env.NODE_ENV === 'test' ? '.' : '../../scripts/'
);
const defaultCollectionFilesDir = path.join(scripts, 'default-files-for-collections');
```

##
_I chose not specify the `scripts/catalog-to-astro-content-directory.js` as an entrypoint because it would result into the following_
```
.
├── ...
├── bin                    
│   ├── dist
│   │   ├── bin
│   │   │   ├── eventcatalog.cjs
│   │   ├── scripts
│   │   │   ├── catalog-to-content-astro-directory.js
├── scripts                    
│   ├── default-files-for-collections
└── ...
```

_What I see with futures PR is_
- development structure
```
.
├── ...
├── astro # astro related files                     
├── default-files-for-collections 
├── src # the scripts dir today
├── eventcatalog.js # the entrypoint to cli
└── ... # other config files
```
- the package in production would be
```
.
├── astro # astro related files                     
├── default-files-for-collections
├── dist # the src build
├── eventcatalog.js # the entrypoint to cli
└── ... # other needed files
```

glc-omid (approved):

boyney123:
Thanks @carlosallexandre , makes sense to me, happy to try it out!

PR #975, opened 2024-11-26
Group astro files under `astro/`

carlosallexandre:
## Motivation
This PR introduces the following improvements:

1. **File Restructure**:
- Grouped all files related to `astro` under the `astro/` directory.
- This is a step towards `pnpm` compatibility where we will split the packages needed by `astro` and install them into users directory.
2. **Publishing Configuration**:
- Removed the `.npmignore` file and switched to using the `files` field in `package.json` to manage publishable files.
- Benefits of this change:
  - Simplifies the configuration by having the list of files to be published directly in the `package.json.`


boyney123:
Thank you @carlosallexandre , really appreciate the PR 

Few ideas / questions

- Can we rename the `astro` folder to `eventcatalog`? Just personal preference really
- Can we move `default-files-for-collections`? They feel somewhat isolated on the root of the project?

Should we move the `catalog-to-astro-content-directory.js` too? The default files are using by this script?
  - is it possible to move them both to new folder? scripts/catalog-to-astro/?
  
 Thoughts?

carlosallexandre:
> * Can we rename the `astro` folder to `eventcatalog`? Just personal preference really

Sure, We can.

> * Can we move `default-files-for-collections`? They feel somewhat isolated on the root of the project?

Usually, assets are placed on the root. And [this comment](https://github.com/event-catalog/eventcatalog/pull/960#issuecomment-2478638700) yet is valid.

With the next PR, on top of that the structure will be something like the following:
- In development
.
├── bin                    
│   ├── eventcatalog.js  # A simple script for import the eventcatalog-cli from `src/`
├── scripts                    
│   ├── test-ci.js
└── ...

bin/eventcatalog.ts
```diff
@@ -94,7 +94,7 @@ program
       const { result } = concurrently([
         {
           name: 'astro',
-          command: 'npm run dev',
+          command: 'npx astro dev',
```
boyney123, line 97: Curious, would this use the latest version of `astro` for users commands?

At the moment from what I undertand, `npm run dev` would use the astro version the catalog is using, these commands may install a different version of astro, and possibly break things ? Although I highly doubt it...? Just wanna know how this works?

carlosallexandre: About that the [npmjs docs](https://docs.npmjs.com/cli/v8/commands/npx#description) says the following:
> Package names provided without a specifier will be matched with whatever version exists in the local project. Package names with a specifier will only be considered a match if they have the exact same name and version as the local dependency.

boyney123: if we continue with `npm run dev` then we have more control in the future with custom scripts etc if we need to use them etc? Do we get much benefit running `npx astro` directly, vs in the package file?

carlosallexandre: > if we continue with npm run dev then we have more control in the future with custom scripts etc if we need to use them etc?

The PR#960 places this custom scripts under the cli command (_which, in my opinion, is a better place_).

> Do we get much benefit running npx astro directly, vs in the package file?

We doesn't have any benefit. But in this PR, the `eventcatalog/` dir doesn't contain a `package.json`. This dir is the dir that will be copied to `.eventcatalog-core/`.

In future PR the `eventcatalog/` will contain a `package.json` similar to this:
```js
{
  "name": "@eventcatalog/astro",
  "version": "1.0.0",
  "author": "event-catalog",
  "private": true,
  "type": "module",
  "scripts": {
    "astro": "astro",
    "build": "astro check --minimumSeverity error && astro build",
    "dev": "astro dev",
    "preview": "astro preview"
  },
  "dependencies": {
    "@astrojs/markdown-remark": "^5.2.0",
    "@astrojs/mdx": "^3.1.5",
    "@astrojs/react": "^3.6.2",
    "@astrojs/tailwind": "^5.1.0",
    "@asyncapi/parser": "^3.3.0",
    "@asyncapi/react-component": "^2.2.2",
    "@changesets/cli": "^2.27.5",
    "@headlessui/react": "^2.0.3",
    "@heroicons/react": "^2.1.3",
    "@stoplight/json-schema-viewer": "^4.7.0",
    "@stoplight/mosaic": "^1.53.2",
    "@tailwindcss/typography": "^0.5.13",
    "@tanstack/react-table": "^8.17.3",
    "astro": "^4.15.4",
    "astro-expressive-code": "^0.36.1",
    "astro-pagefind": "^1.6.0",
    "astro-seo": "^0.8.4",
    "dagre": "^0.8.5",
    "diff": "^7.0.0",
    "diff2html": "^3.4.48",
    "html-to-image": "^1.11.11",
    "lodash.merge": "4.6.2",
    "mermaid": "^10.9.1",
    "prismjs": "^1.29.0",
    "rapidoc": "^9.3.4",
    "react": "^18.3.1",
    "react-dom": "^18.3.1",
    "reactflow": "^11.11.4",
    "rehype-slug": "^6.0.0",
    "remark-gfm": "^3.0.1",
    "semver": "7.6.3",
    "tailwindcss": "^3.4.3",
    "unist-util-visit": "^5.0.0",
    "uuid": "^10.0.0"
  },
  "devDependencies": {
    "@astrojs/check": "^0.9.3",
    "@types/dagre": "^0.7.52",
    "@types/diff": "^5.2.2",
    "@types/lodash.merge": "4.6.9",
    "@types/node": "^20.14.2",
    "@types/react": "^18.3.3",
    "@types/react-dom": "^18.3.0",
    "@types/semver": "^7.5.8",
    "typescript": "^5.4.5"
  }
}
```

With that `package.json` we can use the `npm run dev` again.

### Why I change from `npm run dev` to `npx astro dev`?
When we run the `npm run dev` we change the `cwd` with `execSync` to the `.eventcatalog-core/` dir which doesn't contain a `package.json` with that command, for now.
We couldn't change the `cwd` in the `execSync` and run the `dev` command from the `@eventcatalog/core` package, but there are some places in the code that rely on the `process.cwd()` rather than `process.env.CATALOG_DIR`. So to keep this PR concise i decide to change for `npx astro ...`.

boyney123: Sure sounds good! Happy to try it. Just a few merge conflicts, once they are fixed, let's get it in <3

PR #987, opened 2024-12-03
refactor/setup-workspaces

carlosallexandre:
## Motivation

Closes #575

### Changes Introduced
1. Config workspaces to split deps needed by the `cli` and `astro`.
    - Splitting the deps we are able to install the `astro` deps on the user's directory. This hoist the deps to `.eventcatalog-core/` making them fully compatible with `pnpm`.
2. Auto dependency install
    - Detects which package manager execute the `cli` command and them if doesn't have `node_modules/` inside the `.eventcatalog-core/` install all deps need by the astro.
3. Update Dockerfile of `test-bin.yml` workflow to run with `pnpm`
    - Ensure that EC works with `pnpm`.
    
  
_With that PR in place we can switch to `pnpm` (#884)_

PR #1012, opened 2024-12-10
refactor(core): setup workspaces

carlosallexandre:
## Motivation

Same as #987.

---
Note:
- Should install deps ephemeral or just add to the project user's directory?
- If the deps would be ephemeral, relay on the presence of `node_modules` in development of the package, turns the installation required always because of the monorepo structure.


boyney123:
shall we close this now @carlosallexandre ? Few merge conflicts :O

PR #1021, opened 2024-12-13
fix(core): watcher segmentation fault

carlosallexandre:
## Motivation

When you run `npm run start:catalog` it gaves the `Segmentation fault` error. This is related with the ignore path on watcher. Changing the ignore path to only `catalogDirectory` works. 

As the `catalogDirectory` is usually within the `projectDirectory` (`.eventcatalog-core/`) we won't have a problem with that change. 

_The ignore path was changed to `**/${catalogDirectory}/!(${projectDirectory})**` before because, the CATALOG_DIR was the root of the package and the PROJECT_DIR was `examples/default/` within the CATALOG_DIR. With the restructure of the package, CATALOG_DIR now refers to the `<root>/eventcatalog/` and PROJECT_DIR to `<root>/examples/default/`._

### DIFF
--- a/src/watcher.js
+++ b/src/watcher.js
@@ -1,5 +1,6 @@
 import watcher from '@parcel/watcher';
 import fs from 'node:fs';
+import path from 'node:path';
 import { mapCatalogToAstro } from './map-catalog-to-astro.js';
 import { rimrafSync } from 'rimraf';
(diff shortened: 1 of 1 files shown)


carlosallexandre:
This error seems not happen anymore. Probably an EC dep update. 
Closing...

Issue #1015, opened 2024-12-11
Allow passing arguments to astro cli


capuns:
### Use Case

Currently, eventcatalog cli wraps astro cli functionality but doesn't support passing additional arguments to the underlying astro command. This limits users who need to utilize specific astro cli options or flags while using eventcatalog.

My specific case/need: I'm using dev-containers to develop locally, which has some problem forwarding localhost exposed ports to the host system. I work around this issue by running `astro dev --host` inside the generated `.eventcatalog-core`, as I cannot set the --host flag on the eventcatalog command.

### Proposed Solution

Users should be able to pass through additional arguments to the astro cli using a standard syntax, similar to other cli tools:
Support passing arguments after a delimiter (e.g., --)
Example: `eventcatalog command [eventcatalog-options] -- [astro-cli-options]`

### Implementation Notes

Need to implement argument parsing that distinguishes between eventcatalog options and pass-through arguments while maintaining backward compatibility with existing commands

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out in the [EventCatalog Discord](https://discord.com/invite/3rjaZMmrAm).

boyney123:
Thanks to @carlosallexandre this should now be possible @capuns in @eventcatalog/core@2.16.5

Let us know how you get on

nicobertolucci:
Hey @carlosallexandre and @boyney123 ,
To run EventCatalog on another port, should I use this command? is this correct?
`npm eventcatalog dev -- --port=8000`

carlosallexandre:
@nicobertolucci 

Indeed, the `npx eventcatalog start -- --port=8000` has a bug... the port is fixed on the command ([check here](https://github.com/event-catalog/eventcatalog/blob/main/src/eventcatalog.ts#L130))

is the `npx eventcatalog dev -- --port=8000` worked?

nicobertolucci:
Yup. `eventcatalog dev` works, but `eventcatalog start` **not**.

boyney123:
Thanks @carlosallexandre , this should now be live in `@eventcatalog/core@2.17.1`

PR #1023, opened 2024-12-13
fix: prioritize command arg port over default value

carlosallexandre:
## Motivation

Fix for the bug mentioned [here](https://github.com/event-catalog/eventcatalog/issues/1015#issuecomment-2542025240)


carlosallexandre:
Close in favor of https://github.com/event-catalog/eventcatalog/pull/1024

Issue #930, opened 2024-11-01
RFC - GitHub / GitLab Integration (Generator) with EventCatalog


boyney123:
Hey folks,

This is a RFC (request for comments) for a generator that can integrate with GitHub or GitLab. There are a few ways to do this, and I was going to document an idea I had about it, and would love to see what you think, if this would be useful or not.

### Overview

The generator will let you generate your EventCatalog from many different repositories you have. Each repository could have it's own `/eventcatalog` folder inside it. This folder can have domains, services or messages inside it. The repository will own it's own eventcatalog folders.

Benefits of this approach:

- Each team can be responsible for their own catalog information
- The catalog information can live next to the code
- EventCatalog can pull in this data and give an overview of your architecture
- Teams are responsible for their versioning.

![github-rfc](https://github.com/user-attachments/assets/2494edda-7e15-4a75-b2a8-919c5bd9ab55)

When running EventCatalog (generators) in dev or build, it will fetch the information from GitHub/GitLab repos (tokens required and added to the config file). 

EventCatalog will then:

- Parse each folder in your repos
- Create an EventCatalog for all your domains, services and messages

Questions still to explore:
- How does versioning work? I guess the repo owners would version there own stuff, and these versions are just pulled into EventCatalog.

---

If you have any ideas or thoughts, I would love to hear them

boyney123:
> One way we can go about this is to distribute the generators configuration, and maybe create another generator for Markdown files. That way if a team wants to use Markdown files, other asyncAPI, or others, you don't force them to go down one route or the other.
> 
> Basically, distributing the "generators" what might be a different feature. And yes, this is similar to what compass does with their definition files, that seems to work ok. That would be following a pull model, so eventcatalog pulls the information from the repos.

Yeh might be interesting to explore this more. Do you have any examples in mind, so I can understand the details abit more?

markallanson:
> tokens required and added to the config file

So long as these can be wired in via environment variables too, this is likely to help us in our journey towards adopting event catalog.

IsmaelMartinez:
> > One way we can go about this is to distribute the generators configuration, and maybe create another generator for Markdown files. That way if a team wants to use Markdown files, other asyncAPI, or others, you don't force them to go down one route or the other.
> > Basically, distributing the "generators" what might be a different feature. And yes, this is similar to what compass does with their definition files, that seems to work ok. That would be following a pull model, so eventcatalog pulls the information from the repos.
> 
[...]

PR #1033, opened 2024-12-16
feat(core): added ability to specify dependencies for catalogs

boyney123:
From #930, added the ability to create catalog `dependencies`. These are resources that a team may need for there catalog, but not necessarily own themself.

By specifying a `dependency` EventCatalog will create these resource types for the teams allowing them to use EventCatalog with resources that are external.


carlosallexandre (commented):
Wouldn't be better if the owner of the catalog create the markdown themselves instead of specifying on the config file?

Issue #1043, opened 2024-12-17
Side nav bar isn't sorted


ryancormack:
### Use Case

The current Nav Sidebar isn't in a sorted order. It would be good if the events (for example) were either groups by Domain, or producer, and then sorted in alphabetical order. Or even just sorted in alphabetical order. In the screenshot, OrderCancelled and OrderConfirmed feel very much like they should be next to each other. It makes parsing the whole sidebar harder to find what you are looking for.

![image](https://github.com/user-attachments/assets/37f082cf-0db9-49a1-8807-396140f171c9)

### Proposed Solution

Order the side bar contents in alphabetical order

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out in the [EventCatalog Discord](https://discord.com/invite/3rjaZMmrAm).

boyney123:
Thanks mate, these should now be sorted in 2.18.5 (released in the next 10min of this comment)

PR #1044, opened 2024-12-17
feat: niave example of ordering sidebar

ryancormack:
Niave approach to fixing #1043 

Opned to show an example of what a fix would look like. There's a better way to fix this rather than sorting in the html. I can fix it better later if it's an acceptable approach

image


boyney123:
Thank you mate, I made some changes over on #1060 so will close this for now! Thanks for the issue and even the PR

PR #1074, opened 2025-01-08
fix/ubiquitous-language

carlosallexandre:
## Motivation

This PR upgrades from `astro@5.0.5` to `astro@5.1.2` to leverage the [bug fix in emulated legacy collections where the entry passed to the getCollection filter function did not include the legacy entry fields](https://github.com/withastro/astro/blob/main/packages/astro/CHANGELOG.md#:~:text=%2312875%20e109002%20Thanks%20%40ascorbic!%20%2D%20Fixes%20a%20bug%20in%20emulated%20legacy%20collections%20where%20the%20entry%20passed%20to%20the%20getCollection%20filter%20function%20did%20not%20include%20the%20legacy%20entry%20fields.). This change allows us to use the `slug` legacy field on the `getCollection` filter function to map the `ubiquitous-language` to the requested `domain`.

### DIFF
--- a/.changeset/wise-swans-lie.md
+++ b/.changeset/wise-swans-lie.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+fix(core): ubiquitous-language matching by domains, also astro upgrade to 5.1.x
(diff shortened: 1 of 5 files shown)


boyney123:
Thank you, LGTM lets go!

PR #1080, opened 2025-01-09
Fix tab initialization

### LINKED ISSUE #1079: Tabs not always rendered in case of multiple usage of <Tabs>
### I tried this:

Create a new catalog and add tabs like the following to **multiple** `.md` files:

```
<Tabs>
  <TabItem title="Tab 1">
    ``sh
    This is the content for tab 1
    ``
  </TabItem>
  <TabItem title="Tab 2">
    ``js
    console.log('This is the content for tab 2');
    ``
  </TabItem>
</Tabs>
```

BTW: The docs https://www.eventcatalog.dev/docs/development/components/components/tabs describes the  `TabItem` attribute name as `label` but it should be `title` like in the example.

Run the catalog with `npm run dev`

### This happened:

Click on an item (event/service/domain) with tabs. -> tabs are rendered correctly
Navigate to another item with tabs. -> tabs are **NOT** rendered correctly (no content, see screenshots)

image

### I expected this:

Click on an item (event/service/domain) with tabs. -> tabs are rendered correctly
Navigate to another item with tabs. ->  tabs are rendered correctly

image

### Is there a workaround?

Refreshing the whole page renders the tabs correct again

### Anything else?

_No response_

### EventCatalog Version

@eventcatalog/core@2.18.7

### Node.js Version

v23.1.0

### Platform(s)

MacOS

### Community Notes

carlosallexandre:
## Motivation

Fix #1079

## Changes 

As we use `ViewTransitions`, the `astro:page-load` event is fired whenever happens a navigation, from view transition or native to the browser. So, it adds an event listener to this event to init the tabs.

### DIFF
--- a/.changeset/wild-books-smash.md
+++ b/.changeset/wild-books-smash.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+fix(core): fixed the tab component rendering issues on load of page


boyney123:
Thank you @carlosallexandre makes sense! nice find.

Looks like to CI/CD is failing? Before I can merge

carlosallexandre:
> Thank you @carlosallexandre makes sense! nice find.
> 
> Looks like to CI/CD is failing? Before I can merge

I'm getting the error on `main` either. Seems something with EC deps. I'll dig deeper

PR #1087, opened 2025-01-10
Rework how we sort versions

otbe:
## Motivation

Just noticed that the sorted versions list was not returned by the new functionality which causes a wrong order here for example: https://demo.eventcatalog.dev/docs/domains/Orders/0.0.3
Screenshot 2025-01-10 at 15 25 55


boyney123:
Thanks @otbe and also thanks for the tests!

PR #1102, opened 2025-01-18
feat/discover-checkbox-latest-only

### LINKED ISSUE #702: On discover page add a filter check box to allow people to only show latest version or not.
image

On the discover page, all versions of all resources are loaded. It would be nice to have a checkbox on the page that only shows the latest versions of the resources if its checked.

### LINKED ISSUE #1100: Different versions of Services, Commands, Queries and Events show up as separate in the Search
### I tried this:

I open up the search for either Services, Commands, Queries or Events.

### This happened:

I then see a list with every Service, Command, Query or Event with every version.

### I expected this:

I expected to find the latest version of either a Service, Command, Query or Event, click on it and then perhaps click on an older version if it's still somehow relevant.

### Is there a workaround?

_No response_

### Anything else?

I'm not sure if this is really a from the POV of the developers, but it contributes to a very cluttered search experience, which get exponentially worse with every new version released.
The different numbering also leads to a sense of inconsistency.

Screenshot 2025-01-15 at 14 03 14

Screenshot 2025-01-15 at 14 02 55

### EventCatalog Version

@eventcatalog/core@2.19.3

### Node.js Version

v20.16.0

### Platform(s)

MacOS

### Community Notes

carlosallexandre:
## Motivation

This PR introduces a new checkbox on the discover pages, allowing users to choose whether to display only the latest versions or all versions.
- The checkbox is **checked by default**, meaning users will initially see only the latest versions and can opt out to display all versions if desired. This behavior aligns with the discussion [here](https://github.com/event-catalog/eventcatalog/issues/1100#issuecomment-2595167186).

By enabling this feature:
- When the checkbox is **checked**, the autocomplete/search functionality for names will only show the latest versions, addressing the issue #1100.
- When the checkbox is **unchecked**, the behavior remains unchanged, displaying all versions.

## Related issues
Close #702
Fix #1100

### DIFF
--- a/.changeset/lovely-bags-exercise.md
+++ b/.changeset/lovely-bags-exercise.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+feat(core): users can now filter by latest version of resources on the discover page
--- a/eventcatalog/src/components/Checkbox.astro
+++ b/eventcatalog/src/components/Checkbox.astro
@@ -1,0 +1,38 @@
+---
+import type { HTMLAttributes } from 'astro/types';
+
+export type Props = HTMLAttributes<'input'> & {
+  id: string;
+  name: string;
+  label: string;
+};
+
+const { id, name, label, class: className, ...attrs } = Astro.props;
+---
+
+<label for={id} class="relative block cursor-pointer select-none pl-5">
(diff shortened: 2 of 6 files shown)


boyney123:
Thank you @carlosallexandre , nice quality of life improvement here for folks

PR #1112, opened 2025-01-24
Remove package files left after revert commit

carlosallexandre:
Rever commit 7559d42

## Motivation

This PR remove package files left after revert commit 7559d42. 
_The package file was introduced with PR #987._ 

This package is causing confusion when updating dependencies. Actually, EC is using only the root `package.json`.


boyney123:
Just catching up, so are we removing package.json file from the eventcatalog folder now? If I understand correctly?

carlosallexandre:
> Just catching up, so are we removing package.json file from the eventcatalog folder now? If I understand correctly?

That's right :+1:

boyney123:
Cool thanks @carlosallexandre , lets try it!

PR #1132, opened 2025-02-10
feat/treeview

### LINKED ISSUE #1065: Group the generated events from the OpenAPI by the owning service
### Use Case

In the UI view http://localhost:3000/docs under topic **Queries ()** the number of queries is now a large list of all queries from all the OpenAPI contracts from all our services.

We have multiple services with there own OpenAPI contract.
In every service we use the same query pattern, therefor the queries have the same name. 
Problem arise when we generate with OpenAPI generator the Queries, the 2nd query will override the first one.

### Proposed Solution

Would it be possible to arrange the queries by the services the contracts belongs to?
queries\{service name a}\{the queries for this service}
queries\{service name b}\{the queries for this service}

### Implementation Notes

_No response_

### Community Notes

carlosallexandre:
## Motivation

It adds another view type for sidebar. Now, the user could choose between `FLAT_VIEW` and `TREE_VIEW`. The `FLAT_VIEW` is the default to keep compatibility with old versions.

## Related issues

Close #1065


boyney123:
Thanks @carlosallexandre I think this is great!

A few things I wonder if we can add?

-  Open all root nodes by default (e.g domains is open by default)

I wonder if getTreeView needs to be cached or not? Like we do other pages etc, for large catalogs, this tree view will be called multiple times or is it a build time thing?

eventcatalog/src/components/SideNav/SideNav.astro
```diff
@@ -0,0 +1,24 @@
...
+
+const currentPath = Astro.url.pathname;
+
+let props;
+const SIDENAV_TYPE = EcConfig?.docs?.sidebar?.type ?? 'FLAT_VIEW';
+if (SIDENAV_TYPE === 'FLAT_VIEW') props = await getCatalogResources({ currentPath });
+else if (SIDENAV_TYPE === 'TREE_VIEW') props = getTreeView({ projectDir: process.env.PROJECT_DIR!, currentPath });
+---
+
+<div {...Astro.props}>
```
boyney123, line 21: curious, what does this do? Spreading the props onto the div?

boyney123:
Also I'm working on a new feature that is moving users/teams away from the docs sidebars into their own views... just a heads up but that shoul dnot change this PR, keep them where they are

carlosallexandre:
> A few things I wonder if we can add?
> 
> * Open all root nodes by default (e.g domains is open by default)

I'll add this.

> I wonder if getTreeView needs to be cached or not? Like we do other pages etc, for large catalogs, this tree view will be called multiple times or is it a build time thing?

Great point! I'll add a caching layer to `getTreeView`.

> Also I'm working on a new feature that is moving users/teams away from the docs sidebars into their own views... just a heads up but that shoul dnot change this PR, keep them where they are

This will be similar to the visualizer page, where `teams`/`users`/`channels` aren't added to the treeview.

boyney123:
Thanks for this!

Few questions, 
image
curious what these hooks / code are doing if you could help me understand?

The previous review these were not there, and think the thing was working great, and looks like it's adding a fair amount of code to the project to maintain, so curious to learn more

boyney123:
Just dived deeper into things, removed some of those hooks I don't think we need some of them, and also refactored some the UI too

eventcatalog/src/components/SideNav/SideNav.astro
boyney123: Yeah makes sense to me, thanks mate

boyney123 (approved):

PR #1169, opened 2025-02-18
refactor(core): migrate `users`/`teams` to Astro new content layer API

carlosallexandre:
## Changes Introduced

1. Move `eventcatalog/src/content/config.ts` to `eventcatalog/src/content.config.ts`.
2. Add `glob` loader to `teams`/`users` collections.
3. Remove `teams`/`users` from the custom hydrate phase ( `src/catalog-to-astro-content-directory.js`)

### DIFF
--- a/.changeset/empty-eyes-yell.md
+++ b/.changeset/empty-eyes-yell.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+refactor(core): migrate `users`/`teams` to Astro new content layer API


eventcatalog/src/pages/docs/teams/[id]/index.astro
```diff
@@ -15,23 +15,16 @@ export async function getStaticPaths() {
...
 
-const { render, ...props } = Astro.props;
-
-const { Content } = await render();
+const props = Astro.props;
+const { Content } = await render(props);
 
 const membersRaw = props.data.members || [];
-const members = await Promise.all(membersRaw.map((m) => getEntry(m)));
+const members = (await Promise.all(membersRaw.map((m) => getEntry(m)))).filter(Boolean);
```
Comment (boyney123) [inline eventcatalog/src/pages/docs/teams/[id]/index.astro:27]: Any reason we have to filter by Boolean here? Are things coming back undefined? or?

eventcatalog/src/content.config.ts
```diff
@@ -252,8 +254,16 @@ const ubiquitousLanguages = defineCollection({
   }),
 });
 
+const projectDirBase = (() => {
+  if (process.platform === 'win32') {
+    const projectDirPath = process.env.PROJECT_DIR!.replace(/\\/g, '/');
```
boyney123, line 259: Just trying to wrap my head around the windows, fix, why is that required? Just curious to learn more!

carlosallexandre: The `glob` Astro loader expects a valid `file://` URL, and Windows-style paths (`C:\\some-path`) use backslashes (`\\`), which are invalid in file URLs. So, we replace `\\` with `/` to ensure compatibility.

eventcatalog/src/content.config.ts
```diff
@@ -252,8 +254,16 @@ const ubiquitousLanguages = defineCollection({
...
+    return projectDirPath.startsWith('/') ? projectDirPath : `/${projectDirPath}`;
```
carlosallexandre, line 260: On Unix systems, absolute paths start with `/`, whereas on Windows, they might start with a drive letter (`C:\\...`). Here, we ensure that the path is properly prefixed with `/` to generate a valid file URL (`file://`) required by the glob astro laoder

eventcatalog/src/pages/docs/teams/[id]/index.astro
Comment (carlosallexandre) [reply eventcatalog/src/pages/docs/teams/[id]/index.astro:27]: If I remember well, this is needed because `getEntry` now resolves `undefined` when not found the entry. So, the filter Boolean fix type issues.

eventcatalog/src/content.config.ts
boyney123: Cool thanks for the explanation!

boyney123:
Thanks @carlosallexandre let's give it ago! 

I did notice this, when I run `verify-build:catalog` locally 

```
09:39:13 [WARN] [glob-loader] No files found matching "users/*.md" in directory ""
09:39:13 [WARN] [glob-loader] No files found matching "teams/*.md" in directory ""
```

I'm wondering if its due to how this project is setup or not, going to merge and test on a proper catalog.

But maybe that is something we should look at too, if its effecting end users, as it may cause some confusion

boyney123:
Yeah seems like a local only issue, just tested lastest build and it was fine.

carlosallexandre:
It seems something that happens between the `@astro/check` for types...
I will investigate either and bring the root cause of that.

carlosallexandre:
@boyney123 

As I suspected, adding the PROJECT_DIR env to the type check command remove the warning.

```ts
// scripts/build-ci.js

#!/usr/bin/env node

// This is used for CI on vercel. Must copy files before building.
import { join } from 'node:path';
import { execSync } from 'node:child_process';
const __dirname = import.meta.dirname;

const args = process.argv.slice(2);
const catalog = args[0] || 'default';

const catalogDir = join(__dirname, '../eventcatalog/');
const projectDIR = join(__dirname, `../examples/${catalog}`);

// Build cli
execSync('pnpm run build:bin', { stdio: 'inherit' });

// Run the generator
execSync(`cross-env NODE_ENV=CI PROJECT_DIR=${projectDIR} CATALOG_DIR=${catalogDir} npx . generate`, {
  stdio: 'inherit',
});

// Build catalog
execSync(`cross-env NODE_ENV=CI PROJECT_DIR=${projectDIR} CATALOG_DIR=${catalogDir} npx . build`, {
  stdio: 'inherit',
});

// Type check
execSync(`pnpm exec astro check --minimumSeverity error --root ${catalogDir}`, {
  stdio: 'inherit',
  env: {
    ...process.env,
    PROJECT_DIR: projectDIR,
    CATALOG_DIR: catalogDir,
  },
});
```

PR #1234, opened 2025-03-07
fix: Icon images now also work when a base path is used

WellingR:
## Motivation

This fixes a bug where icons were not being loaded when a base path was configured.

I tested this by configuring a `base` path in the default example and by running `pnpm run start:catalog`

Not sure how to create an automated test for this...


carlosallexandre (approved):

boyney123:
Thanks again @WellingR !

PR #1235, opened 2025-03-07
fix: Redirects to latest version now work correctly when basePath is used

WellingR:
## Motivation

Links on the new architecture screen ("View in visualizer" and "Read documentation") link to a service/domain without verion. The redirect code was broken as it would duplicate the base path (which would generate an incorrect link)

This PR fixes that issue

### DIFF
--- a/eventcatalog/src/pages/visualiser/[type]/[id]/index.astro
+++ b/eventcatalog/src/pages/visualiser/[type]/[id]/index.astro
@@ -32,8 +32,7 @@ const pageTitle = `${props.collection} | ${props.data.name}`.replace(/^\w/, (c) 
 
 const { pathname } = Astro.url;
 
-// redirect with any search params too
-const redirectUrl = buildUrl(pathname + '/' + props.data.latestVersion);
+const redirectUrl = buildUrl(pathname + '/' + props.data.latestVersion, false, true);
 ---
 
 <!doctype html>
@@ -48,6 +47,7 @@ const redirectUrl = buildUrl(pathname + '/' + props.data.latestVersion);
 
 <script define:vars={{ redirectUrl }}>
   document.addEventListener('DOMContentLoaded', () => {
+    // redirect with any search params too
     const searchParams = window.location.search;
     const fullRedirectUrl = redirectUrl + searchParams;
     window.location.href = fullRedirectUrl;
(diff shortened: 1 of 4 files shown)


eventcatalog/src/pages/visualiser/[type]/[id]/index.astro
```diff
@@ -32,8 +32,8 @@ const pageTitle = `${props.collection} | ${props.data.name}`.replace(/^\w/, (c)
 
 const { pathname } = Astro.url;
 
-// redirect with any search params too
-const redirectUrl = buildUrl(pathname + '/' + props.data.latestVersion);
+// pathname already includes the base path
+const redirectUrl = pathname + '/' + props.data.latestVersion;
```
Comment (carlosallexandre) [inline eventcatalog/src/pages/visualiser/[type]/[id]/index.astro:36]: What happens if the user config the trailing slash as true??? The redirects still works?

carlosallexandre (changes requested):

boyney123:
@WellingR thanks for this.

Can you let me know how I can replicate the issue? Just had a local catalog with the trailing slash to true, build the catalog and run preview, but the links seem to work ok?

eventcatalog/src/pages/visualiser/[type]/[id]/index.astro
Comment (WellingR) [reply eventcatalog/src/pages/visualiser/[type]/[id]/index.astro:36]: Good point, I will do some additional tests with the trailing slash feature enabled

Comment (WellingR) [reply eventcatalog/src/pages/visualiser/[type]/[id]/index.astro]: Thanks for catching this, this has been fixed int he latest commit

WellingR:
@boyney123 

The following gif shows the bug. 

To reproduce configure the default "FlowMart" catalog with
```
  base: '/testing',
  trailingSlash: true,
```

and run `pnpm run start:catalog` to start the catalog

Eventcatalog_bug

carlosallexandre (approved):
LGTM! 🚀

One suggestion: Should the second argument of buildUrl be an options object? If we keep adding parameters, it might get messy over time.

That said, this change seems unrelated to the main focus of this PR—maybe it would be better to address it in a separate one.

Great work! 👍

boyney123:
Thanks @WellingR @carlosallexandre lets get it in!

boyney123:
@all-contributors add @WellingR for code

PR #1258, opened 2025-03-17
fix(examples): Fix PlaceOrder command definition

vienin:
* Move folder InventoryService/commands/PlaceOrder to OrdersService/commands/PlaceOrder

* Add PlaceOrder command to the OrdersService received messages

## Motivation

While using the demo app, I realized that the PlaceOrder command was not associated to any service. Reading the flow and an example sequence diagram in the Orders domains, I figured out that the PlaceOrder command should be associated as the received commands on the OrdersService.

Hoping the fix is correct.


boyney123:
Thank you @vienin makes sense to me! Nice find, and thanks for spending the time to make the PR

vienin:
> Thank you @vienin makes sense to me! Nice find, and thanks for spending the time to make the PR

You'r welcome, thanks for the merge !

Issue #1238, opened 2025-03-10
Don't auto-expand services in SideNav/TreeView


VitaliiBalash:
### Use Case

Hello.
The current implementation of TreeView automatically expands domains, services, and channels (see [getTreeView.ts#L119](https://github.com/event-catalog/eventcatalog/blob/main/eventcatalog/src/components/SideNav/TreeView/getTreeView.ts#L119)). Although this makes sense for domains, it can diminish the user experience when a domain contains a large number of services, as it becomes harder to view and access other components within that domain.

### Proposed Solution

It can be approached in different ways:

1. remove `service` from the auto-expansion list
2. only auto-expand if there are less than X services
3. make the list [getTreeView.ts#L119](https://github.com/event-catalog/eventcatalog/blob/main/eventcatalog/src/components/SideNav/TreeView/getTreeView.ts#L119) configurable via `eventcatalog.config.js`

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out in the [EventCatalog Discord](https://discord.com/invite/3rjaZMmrAm).

PR #1259, opened 2025-03-17
feat(core): collapse everything except domains in TreeView

VitaliiBalash:
- collapses everything except domains in TreeView 

## Motivation

Hey there! I really appreciate you for making such a fantastic tool!

The current implementation of TreeView automatically expands domains, services, and channels (see [getTreeView.ts#L119](https://github.com/event-catalog/eventcatalog/blob/main/eventcatalog/src/components/SideNav/TreeView/getTreeView.ts#L119)). Although this makes sense for domains, it can diminish the user experience when a domain contains a large number of services, as it becomes harder to view and access other components within that domain.

I've chosen the simplest solution for now, but I'm happy to explore other approaches if you're generally open to the idea.

Alternative solutions:

- only auto-collapse if there are more than X services
- make the list [getTreeView.ts#L119](https://github.com/event-catalog/eventcatalog/blob/main/eventcatalog/src/components/SideNav/TreeView/getTreeView.ts#L119) configurable via eventcatalog.config.js

### DIFF
--- a/.changeset/little-tables-pick.md
+++ b/.changeset/little-tables-pick.md
@@ -1,0 +1,5 @@
+---
+'@eventcatalog/core': patch
+---
+
+feat(core): collapse everything except domains in TreeView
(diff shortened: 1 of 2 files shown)


VitaliiBalash:
Related to https://github.com/event-catalog/eventcatalog/issues/1238

boyney123:
Makes sense to me, lets go for it, if others provide feedback we can review this and maybe add config in `eventcatalog.config.js` file.

Issue #1274, opened 2025-03-19
Improve ⌘K search results


VitaliiBalash:
### Use Case

Improve ⌘K search results

<img width="285" alt="Image" src="https://github.com/user-attachments/assets/6d1abc8d-e59e-4ecc-a00d-7ac1648fe722" />

<img width="277" alt="Image" src="https://github.com/user-attachments/assets/db21331e-1a79-42b2-80fb-e3225a785a59" />

### Proposed Solution

There are several approaches that can be implemented to enhance the quality of the search results.

- Use Service/Message name as a title

<img width="191" alt="Image" src="https://github.com/user-attachments/assets/c7785bfd-dedb-4350-9ecf-2119671f329a" />

- Don't index Service/Domain/...SideBar. 

Although viewing related pages in the search results can be helpful, I personally find it simpler to go directly to the specific entity and then navigate from there. Additionally, as more elements are added, the search results will become increasingly cluttered.

- Don't index not relevant pages

Removing architecture, changelog, docs/[type]/index.astro (redirect to the versioned page) and other irrelevant pages can greatly improve search accuracy.

<img width="175" alt="Image" src="https://github.com/user-attachments/assets/8c2c502f-cc35-4c49-a303-5ecf4a892626" />

### Implementation Notes

Add pagefind tags to the page templates

- `data-pagefind-meta="title"` can be used to explicitly tell where pagefind should take the title from
- `data-pagefind-ignore` to ignore certain blocks, like SideBars
- `data-pagefind-body` can be used to limit the indexing scope on a page. Also, once added, any page that doesn't include it will be ignored.

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out in the [EventCatalog Discord](https://discord.com/invite/3rjaZMmrAm).

VitaliiBalash:
I'm happy to open a pull request if you're open in any of these suggestions 😉

boyney123:
Hey @VitaliiBalash 

Sounds great, and a nice improvement, yeah if you want to make a PR why not! 

I also do wonder if we should have `pagefind` attribute like this too? https://starlight.astro.build/reference/frontmatter/#pagefind, on any resource? So the whole thing can be ignored if folks choose to do so? Any thoughts on this?

[...]

PR #1278, opened 2025-03-20
feat(core): improve search quality

VitaliiBalash:
## Motivation

Can be found here https://github.com/event-catalog/eventcatalog/issues/1274

## Changes

- use page title as search result title
- ignore Domain/Service/...SideBar
- index only the documentation pages, all other are ignored
  - `docs/`
    - `[type]/[id]/`
      - `[version]/`
        - `asyncapi/index.astro`
        - `spec/index.astro`
        - `index.astro`
      - `language/[dictionaryId]/index.astro`
      - `language.astro`
    - `teams/[id]/index.astro`
    - `users/[id]/index.astro`
- index only the latest version
- change titles for OpenAPI and AsyncAPI pages. Example: `AsyncApi Spec` -> `Services | Service Name | AsyncApi Spec`

Screenshot 2025-03-20 at 16 04 36

## Notes

I tried to improve search accuracy by manually adjusting the header weights using `data-pagefind-weight="10"`, but it didn’t seem to make a difference. I must be missing something.

I'm not sure how valuable it is to index OpenAPI and AsyncAPI spec pages, but I’ve included them just in case.

I’ve excluded changelogs and old resource versions from the index. If you think they’d be useful in search results, what about:
- adding a configuration option in `eventcatalog.config.js`
- using filters ([Pagefind filtering](https://pagefind.app/docs/filtering/)), though that would require improvements to the search UI.

Screenshot 2025-03-20 at 16 27 16


VitaliiBalash:
My bad, didn’t run the linter locally 🤦‍♂️ Will fix tomorrow.

boyney123:
Thank you for this @VitaliiBalash , great PR and overview, appreciate that, helped me get up to speed.

I'm happy with this, we put pagefind is as it was, and we have had various feedback about the use of it (noise). So think this is a great baseline for it going forward.

Let's see what community come back with, if we get feedback, we can add more as times goes on like previous versions etc etc.

One thing that would be nice... not sure how to do yet, would be to let users search there specs better. Like the operationIds or something like this.... but we can look at that in the future.

Thanks again for your time and the PR, appreciate it.

PR #1281, opened 2025-03-21
Fix dev crashing in devcontainers on change

Jessewb786:
## Motivation

If you run eventcatalog in a devcontainer and run the following script:

eventcatalog dev -- --host

Then upon editing a file in the eventcatalog, the dev instance of eventcatalog will crash.
This is a known problem in astro as seen here: https://github.com/withastro/docs/issues/4003

## Solution

Add to the vite configuration in astro.config.mjs and this PR contributes this first part.

```javascript
server:{
    host: "0.0.0.0",
    hmr: { clientPort: <port> },
    port: <port>, 
    watch: { usePolling: true }
}
```

The second part of the solution cannot be done in eventcatalog, which is including the following in the .devcontainer file:

```
"forwardPorts": [4321, <port>],
"containerEnv": {
    "CHOKIDAR_USEPOLLING": "true"
} 
```

### DIFF
--- a/.changeset/tricky-walls-buy.md
+++ b/.changeset/tricky-walls-buy.md
@@ -1,0 +1,5 @@
+---
+'@eventcatalog/core': patch
+---
+
+Fix dev crashing in devcontainers on change
--- a/eventcatalog/astro.config.mjs
+++ b/eventcatalog/astro.config.mjs
@@ -16,11 +16,12 @@ import expressiveCode from 'astro-expressive-code';
 
 const projectDirectory = process.env.PROJECT_DIR || process.cwd();
 const base = config.base || '/';
+const port = config.port || 3000;
 
 // https://astro.build/config
 export default defineConfig({
   base,
-  server: { port: config.port || 3000 },
+  server: { port: port },
 
   outDir: config.outDir ? join(projectDirectory, config.outDir) : join(projectDirectory, 'dist'),
 
@@ -68,6 +69,12 @@ export default defineConfig({
     },
     worker: {
       format: 'es',
+    },
+    server:{
+      host: "0.0.0.0",
+      hmr: { clientPort: port },
+      port: port,
+      watch: { usePolling: true }
     },
     build: {
       commonjsOptions: {


boyney123:
Thank you @Jessewb786.

I'm wondering if it makes sense to introduce a new config value in `eventcatalog.config.js` for this (or not)? 
For example some things we enable here like `fsWatch` have this

```
/**
   * Whether to use fs.watchFile (backed by polling), or fs.watch. If polling leads to high CPU
   * utilization, consider setting this to `false`. It is typically necessary to **set this to
   * `true` to successfully watch files over a network**, and it may be necessary to successfully
   * watch files in other non-standard situations. Setting to `true` explicitly on OS X overrides
   * the `useFsEvents` default.
   */
  usePolling?: boolean
  ```
  
  I'm not sure what default settings are with Astro, so I have to look, but might be easier if we introduce a new setting that will set this (server value), any thoughts on that?

Jessewb786:
@boyney123
Currently I change the dev script from 'eventcatalog dev' to 'eventcatalog dev -- --host' and included this change.

The advantage of this current setup is that it seems to work for both devs using .devcontainers and those running the script natively.

However, perhaps the most convenient solution, for this particular case, would be to detect if one is running inside a devcontainer or not and then configuring astro accordingly.
I'm not sure what the best way of doing that is, but there are candidate possibilities:

- The .dockerenv file at the root of the filesystem
- Use one of the environment variables that consistently appear in devcontainers.

If that's not desired, then indeed, just setting a configuration variable in eventcatalog.config.js would be the next best thing if done in a way that doesn't require modifying the script.

Or depending on the desired flexibility you would like to provide, let general astro configuration be done directly via eventcatalog.config.js and pass it on so that the eventcatalog itself doesn't have to be updated every time someone finds the need to edit astro's configuration.

boyney123:
Sorry been awhile here on this issue, is this still a problem that needs addressing @Jessewb786 ?

boyney123:
Will close this for now, please re open if its an issue.

Issue #1317, opened 2025-04-01
Autogenerated (custom docs) does not work if children are folders (not files)


boyney123:
### Use Case

Here is an example

<img width="1336" alt="Image" src="https://github.com/user-attachments/assets/92a239d7-fbcb-4959-acf0-fc9ed0c61896" />

We are trying to tell EventCatalog to autogen this folder, but the children themselfs are folders, EventCatalog expects children to be files for a folder.

### Proposed Solution

EventCatalog should allow folder children and just handle them for the users. In this case the children would be sub menu items in the sidebar, and the config should be valid.

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out in the [EventCatalog Discord](https://discord.com/invite/3rjaZMmrAm).

boyney123:
Now live

PR #1321, opened 2025-04-01
feat: support directories inside autogenerated custom docs

callingmybluff:
## Motivation
I've encountered bug #1317 and this is my attempt to solve it.

The only "known issue" is that folders inside the autogenerated folders should have a slug-like naming
* No capital letters
* No spaces (I believe), but rather dashes.

## Desclaimer
I have not written JS code in over 2 years, so any comments and review is encouraged


boyney123:
Thank you @callingmybluff I really appreciate the effort here and improvement. I reviewed the code, took parts of it, but was easier to make a new PR https://github.com/event-catalog/eventcatalog/pull/1336, with all the changes.

Will close this one, but thanks for making the time, and making the custom docs even better!

PR #1330, opened 2025-04-02
feat(core): fixes and improvements for ResourceLink

VitaliiBalash:
## Changes

- (fix) Avoid trailing and leading whitespaces. Trailing underline appears when text is placed directly after the component.
```
<ResourceLink type="domain" id="Orders">Orders</ResourceLink> test
```
Screenshot 2025-04-02 at 19 42 32
Screenshot 2025-04-02 at 19 42 56

- (fix|improvement) When no version is specified, point to the latest version instead of `/docs/${collection}/${id}`. 
  - The redirect pages are not generated for flows and queries ([source](https://github.com/event-catalog/eventcatalog/blob/main/eventcatalog/src/pages/docs/%5Btype%5D/%5Bid%5D/index.astro#L34)). The current implementation just points to a non existing page. 
  - This also avoids the needless intermediate redirect 
- (improvement) Fallback to resource name if no slot content provided.  I’d assume in most cases users will want to use the name, not custom text. By the way, the example in the [documentation](https://www.eventcatalog.dev/docs/development/components/components/resource-link) `<ResourceLink id="InventoryService" type="service" />` is a bit confusing - as-is, it produces no visible link 😄 
- (improvement) Log if failed to resolve the resource. Will help to catch mistakes in documentation. I wonder if it would be better to just fail and force the author to correct the mistake right away.

### DIFF
--- a/eventcatalog/src/components/MDX/ResourceLink/ResourceLink.astro
+++ b/eventcatalog/src/components/MDX/ResourceLink/ResourceLink.astro
@@ -8,6 +8,7 @@ const collection = resourceToCollectionMap[type as keyof typeof resourceToCollec
 
 let slotHTML = await Astro.slots.render('default');
 let href = '#';
+let linkHasError = false;
 
 try {
   if (collection === 'users' || collection === 'teams') {
@@ -18,6 +19,7 @@ try {
     href = buildUrl(`/docs/${collection}/${id}`);
   } else {
     const resourcesCollection = await getCollection(collection);
+
     const resources = getItemsFromCollectionByIdAndSemverOrLatest(resourcesCollection, id, version);
     const resource = resources[0];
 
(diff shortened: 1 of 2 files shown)


boyney123:
Great thank you for for the improvement @VitaliiBalash 

I also added a UI state if there are broken links so people can quickly see something isnt quite right.

I will also update the docs to reflect the changes. Thanks for taking the time to fix this!

eventcatalog/src/components/MDX/ResourceLink/ResourceLink.astro
```diff
@@ -1,23 +1,50 @@
...
   )
 }
 
 {
-  !version && (
-    <a href={buildUrl(`/docs/${collection}/${id}`)} class="text-purple-500 hover:text-purple-700">
-      <slot />
+  !linkHasError && (
+    <a href={href} class="text-purple-500 hover:text-purple-700">
+      {slotHTML}
```
VitaliiBalash, line 47: If the content inside <a> is on a separate line (like `<a>\n\t{content}\n</a>`), it ends up adding a trailing whitespace, which causes an unwanted underline at the end. There might be a better way to handle or prevent this.

boyney123: Ah OK, do you want me to hold off merging?

VitaliiBalash: Ignore the comment. There is no trailing underline this time 😮‍💨 
Screenshot 2025-04-03 at 11 49 55

boyney123: Cool, shall we merge?

VitaliiBalash: Yes, please.
Honestly, I have no clue why the issue was happening in the first place 🤦

boyney123: Yeah no problem, thanks again for taking the time to improve for everyone!

VitaliiBalash:
Thank you for the review!

PR #1354, opened 2025-04-07
feat(core): added experimental api for catalog

boyney123:
Added experimental API for EventCatalog, this `/api/catalog` route will return a dump of the EventCatalog.

API is experimental at the moment, whilst I explore other tools/integrations for EventCatalog.

API will be stable if things work out as expected. For now you have to opt into this.


carlosallexandre:
Great to see the use of `@eventcatalog/sdk` here. 
_Would be great to get all stuff to build the catalog with the sdk either rather having custom code to get..._

boyney123:
_Would be great to get all stuff to build the catalog with the sdk either rather having custom code to get..._ yeah interesting idea. I worry about moving away from Astro Collections through and native stuff the framework gives us in this regard, as I think it gives loads of performance benfits etc, do you mean move away from that stuff into more SDK stuff?

PR #1375, opened 2025-04-09
fix(core): fix versions list in changelog

VitaliiBalash:
## Motivation

When rendering changelog pages, the version list includes versions from all items in the collection, rather than only the item the changelog is being generated for. This causes the previous version to be calculated incorrectly, resulting in a missing diff between the current and the actual previous version.

For example:
- the resource for which the changelog is being generated has versions 2.0.0 and 1.0.0.
- there is another resource in the same collection with version 1.5.0.

In this case, the system incorrectly identifies 1.5.0 as the previous version instead of 1.0.0, resulting in the diff being calculated between 2.0.0 and 1.5.0.

### DIFF
--- a/.changeset/flat-days-smile.md
+++ b/.changeset/flat-days-smile.md
@@ -1,0 +1,5 @@
+---
+'@eventcatalog/core': patch
+---
+
+fix(core): fix versions list in changelog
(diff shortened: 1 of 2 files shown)


boyney123:
Ah yeah makes sense to me, let's get it in!

Thanks again @VitaliiBalash for the contribution

PR #1380, opened 2025-04-15
fix(core): use the node label in simple mode too

VitaliiBalash:
## Motivation

As of now, the node label is used only in the full mode. This PR add is to the simple mode.


boyney123:
Thank you @VitaliiBalash 🙏

PR #1399, opened 2025-04-16
fix(core): style of sidebar labels

VitaliiBalash:
## Motivation

A small change to make style of sidebar labels consistent.
-  There is no need to have resource type in the label. If we are on a Service page, it is obvious that Owners stands for service owners. What we have right now:
   - ServiceSidebar: `Versions` but `Service Owners`
   - DomainSidebar: `Services` and `Versions` but `Domain Owners`
   - MessageSidebar: `QueryVersions`, `Query Owners`, ...
   - etc... 
- Make RSS and Channel protocols semibold to match other labels

### DIFF
--- a/.changeset/mean-lemons-rescue.md
+++ b/.changeset/mean-lemons-rescue.md
@@ -1,0 +1,5 @@
+---
+'@eventcatalog/core': patch
+---
+
+fix(core): style of sidebar labels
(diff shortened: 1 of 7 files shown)


boyney123:
Thank you @VitaliiBalash makes sense to me, thanks for this nice improvement.

PR #1400, opened 2025-04-16
fix(core): propagate vite error in eventcatalog build correctly

alexanderhorner:
## Motivation

Our gitlab pipelines showed as successful even though the eventcatalog build failed. The problem lies with the unix command executed with `execSync`. It's a small fix, just adding `bash -c "set -o pipefail;`

### DIFF
--- a/.changeset/unlucky-numbers-invent.md
+++ b/.changeset/unlucky-numbers-invent.md
@@ -1,0 +1,5 @@
+---
+'@eventcatalog/core': patch
+---
+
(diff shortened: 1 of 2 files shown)


boyney123:
Hey @alexanderhorner 

Thanks for fixing this, nice find. Happy to add this one, thanks for the contribution too!

boyney123:
@all-contributors add @alexanderhorner for code

boyney123 (approved):

Issue #575, opened 2024-07-11
Issue building empty template project with pnpm package manager


mortezaalizadeh:
Hi,

I created a event-catalog project using the command npx @eventcatalog/create-eventcatalog@latest my-catalog. The build process works perfectly with "npm run build", but I'm encountering an error when switching to pnpm running "pnpm build" command. Running "pnpm dev" results in the same error.

Does anyone know how to successfully build an event-catalog project with pnpm?

Thanks!

`
╭─morteza@pop-os ~/tmp/my-catalog  ‹main*› 
╰─➤  pnpm build                                                                                                                                                                                                                                                                                                                                                                                                                     1 ↵

> my-catalog@0.1.0 build /home/morteza/tmp/my-catalog
> eventcatalog build

Building EventCatalog...

> @eventcatalog/core@2.0.12 build
> npm run scripts:hydrate-content && astro check --minimumSeverity error && astro build


> @eventcatalog/core@2.0.12 scripts:hydrate-content
> node scripts/catalog-to-astro-content-directory.js

node:internal/modules/run_main:129
    triggerUncaughtException(
    ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'glob' imported from /home/morteza/tmp/my-catalog/.eventcatalog-core/scripts/catalog-to-astro-content-directory.js
Did you mean to import "glob/dist/commonjs/index.js"?
    at packageResolve (node:internal/modules/esm/resolve:854:9)
    at moduleResolve (node:internal/modules/esm/resolve:927:18)
    at defaultResolve (node:internal/modules/esm/resolve:1157:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:383:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:352:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:227:38)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:87:39)
    at link (node:internal/modules/esm/module_job:86:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}

Node.js v20.15.0
node:child_process:965
    throw err;
    ^

Error: Command failed: PROJECT_DIR='/home/morteza/tmp/my-catalog' CATALOG_DIR='/home/morteza/tmp/my-catalog/.eventcatalog-core' npm run build
    at genericNodeError (node:internal/errors:984:15)
    at wrappedFn (node:internal/errors:538:14)
[...]

PR #1423, opened 2025-04-22
(WIP): fix pnpm compatibility

carlosallexandre:
## Motivation

https://github.com/event-catalog/eventcatalog/issues/575

As we rollback the monorepo setup that fixed the pnpm compatibility because the changeset, this PR brings it again working with changeset.

_Note: This is WIP_


boyney123:
Thanks for your effort on this, @carlosallexandre – I really appreciate the time you've put into it.

At the moment, I don't have the capacity to review and merge this, as I’m focusing on other priorities. Ensuring everything works smoothly without breaking the existing model will take some time, so I’ll need to put this on hold for now.

EC's current setup is working well and allows me to maintain a lean and sustainable approach, so I’d like to revisit this in a month or two when I can give it the attention it deserves.

Hope that makes sense, and thanks again for your contribution!

carlosallexandre:
Thanks for the response. Just keeping this open for visibility and future reference. Let me know if you’d prefer me to close it.

boyney123:
Tidying up on issues and PR, going to close this one for now 👌

SebJones:
Is there any appetitie to revisit this? We're trying to set up inside a pnpm workspaces monorepo and would love this to work out of the box

boyney123:
@SebJones yes there is, think we have to figure out how to enable this without too much core refactoring...

carlosallexandre:
The pnpm compatibility issue happens because EventCatalog copies `eventcatalog/` into `.eventcatalog-core/`, which breaks package resolution.

Possible solutions without a major refactor:
1. **Avoid creating `.eventcatalog-core/`**
Instead, resolve the `eventcatalog/` package directly from the user directory using Astro collections.
The downside is serving static assets — if I remember correctly, this would require a custom server (Astro might support this).

2. **Extract dependencies for .eventcatalog-core/**
My approach was using a monorepo so we explicitly control which dependencies the catalog needs.
Another option would be scanning `eventcatalog/` to detect used dependencies, but this is fragile, especially with dynamic imports.

    The flow would be:
    - Detect required dependencies
    - Generate a `package.json` (or similar) for `.eventcatalog-core/`
    - Install dependencies after creating `.eventcatalog-core/`
---
I haven’t touched this code in a long time, so I can’t guarantee either solution will work as-is, but this should point in the right direction.

PR #1449, opened 2025-05-07
fix(core): fix protocol icon nullability issue

XaaXaaX:
## Motivation
The protocol icon caused graphe being empty , giving following error in browser console 
Screenshot 2025-05-07 at 17 06 49

The issue is happening only if a channel does not provide protocols property in markdown


boyney123:
Thank you mate!

boyney123 (approved):

PR #1465, opened 2025-05-16
feat(core): added subdomain folder support

boyney123:
Added the ability to create `subdomains` folder.

Example: `domains/MyDomain/subdomains/MySubdomain/index.mdx`

You can also choose not to have subdomain folders if you prefer that.


boyney123:
Closes #1465

PR #1470, opened 2025-05-18
Allow channels to be displayed as a single node (Fix #989)

### LINKED ISSUE #989: Channels are duplicated in the domain view
### I tried this:

I created a channel, to which I added several events

### This happened:

When viewing the domain, all events are attached to a dedicated channel instance. Not only is this wrong, to my understanding, but this drastically affects performances.

### I expected this:

When viewing the domain, all events should be attached to the SINGLE channel instance they belong to.

### Is there a workaround?

_No response_

### Anything else?

_No response_

### EventCatalog Version

_No response_

### Node.js Version

_No response_

### Platform(s)

_No response_

### Community Notes

ababilone:
## Motivation

Allow channels node to be displayed as a single node per configuration (Fixing #989)

### DIFF
--- a/.changeset/salty-wombats-brake.md
+++ b/.changeset/salty-wombats-brake.md
@@ -1,0 +1,5 @@
+---
+'@eventcatalog/core': minor
+---
+
+Allow channels to be displayed as a single node (Fix #989)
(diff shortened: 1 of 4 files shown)


boyney123:
Thank you @ababilone , LGTM

I just updated the config slightly, added `visualiser` namspace to it, in case we have future config vars for the visualiser too.

PR #1502, opened 2025-05-28
feature(core) Add server host binding property

t0kieu:
## Motivation

Add server host binding property, for correctly staring nodejs/astro server with specific IP defined in configuration file.

### DIFF
--- a/.astro/types.d.ts
+++ b/.astro/types.d.ts
@@ -1,1 +1,2 @@
 /// <reference types="astro/client" />
+/// <reference path="content.d.ts" />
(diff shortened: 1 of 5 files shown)


boyney123:
Thanks @t0kieu , this actually introduces some breaking changes to EventCatalog, moving the properties to the `server` property and also defaulting to `0.0.0.0` astro defaults host to `false`. 

I made a non breaking verison here https://github.com/event-catalog/eventcatalog/pull/1510.

Thanks for the PR and idea, will close this one and progress with the non breaking version. Thanks!

PR #1560, opened 2025-06-23
feat(core): added new ProtoViewer componenot

boyney123:
New ProtoViewer componet that will take a proto file and render it like this.

### Usage
```
<ProtoViewer file="x"/>
```

image


boyney123:
Tidying up on issues and PR, going to close this one for now 👌

PR #1591, opened 2025-07-10
(feat) add ability to restrict which entities are shown on the Entity…

simonwfarrow:
…Map by  passing in an array of valid Entity Ids i.e <EntityMap id="domain-id" entities="{['entityA','entityB']}"/>

## Motivation

The Node Graphs can become quite large. On a Service page for example I wish to show a limited set of Entities that the service works with rather than all the Entities that are part of the domain

### DIFF
--- a/eventcatalog/tsconfig.json
+++ b/eventcatalog/tsconfig.json
@@ -19,6 +19,7 @@
     },
     "jsx": "react-jsx",
     "jsxImportSource": "react",
-    "types": ["vitest/globals"]
+    "types": ["vitest/globals"],
+    "lib": [ "dom", "dom.iterable", "es2016", "esnext" ]
   }
 }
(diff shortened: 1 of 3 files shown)


eventcatalog/tsconfig.json
```diff
@@ -19,6 +19,7 @@
     },
     "jsx": "react-jsx",
     "jsxImportSource": "react",
-    "types": ["vitest/globals"]
+    "types": ["vitest/globals"],
+    "lib": [ "dom", "dom.iterable", "es2016", "esnext" ]
```
boyney123: Hey @simonwfarrow 

Thanks for the PR.

Any reason we had to change the tsconfig?

boyney123 (changes requested):

boyney123:
Thanks for this @simonwfarrow this makes sense to me, will review!

boyney123:
@all-contributors add @simonwfarrow for code

boyney123:
Thanks @simonwfarrow , docs have been updated and the feature will be in 2.51.1

simonwfarrow:
Thank you!

PR #1632, opened 2025-07-28
fix(core): Fix for issue-1631. Reverts removal of ClientRouter that d…

SodaDev:
## Motivation

This PR fixes issues-1631


boyney123:
thanks for this, I believe this will be fixed in our next major version v3 , will re-open if we have problems.

PR #1637, opened 2025-07-28
feat(core): update search to custom pagefind solution

boyney123:
Update to search for a custom pagefind solution, rather than astro-pagefind.

• Allows people to filter by resource type
• Do exact matching on search terms
• Index moved to build step (outside of astro).

Screenshot 2025-07-28 at 15 51 34


boyney123:
@DenBeke Hopefully this is a good solution for you guys, improved search

PR #1803, opened 2025-11-04
Update vulnerable dependencies

vladajankovic:
## Motivation

While exploring and playing around with EventCatalog, I noticed that it had some package vulnerabilities (one was even critical level). A small contribution but it's always important to resolve critical vulnerabilities in code or in dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -156,5 +156,16 @@
     "vite-tsconfig-paths": "^4.3.2",
     "vitest": "2.1.9"
   },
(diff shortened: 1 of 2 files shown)


boyney123:
Thanks you @vladajankovic , really appreciate the time to do this. 

This new version of vitest broke some tests, I had to fix them over at #1806

Will close this one, as it's now implemented over there.

Thanks for the idea/contribution, everything helps!

Issue #1799, opened 2025-10-28
Add support for viewing Avro (.avsc) schemas in SchemaViewer


mumundum:
### Use Case

We use **Avro (`.avsc`)** as the authoritative source of truth for our event payload schemas.  
Today we render these schemas in EventCatalog using the `<Schema />` component, which works well for displaying the raw schema and providing download access.

However, the `<SchemaViewer />` component offers a significantly better experience for developers and non-technical stakeholders — including collapsible field navigation, search, and clearer visual structure. This is extremely helpful for understanding complex event payloads, onboarding new engineers, and cross-team communication.

Since `<SchemaViewer />` currently supports **JSON Schema only**, teams using Avro are limited to plain text schema display and lose the benefits of the more user-friendly viewer UI.

We would like to avoid maintaining duplicated schemas or manual conversion pipelines. Ideally, EventCatalog would allow Avro schemas to benefit from the same interactive viewer experience as JSON Schema.

### Proposed Solution

Allow `<SchemaViewer />` to handle **Avro (`.avsc`)** schemas.

This could be achieved in one of two ways:

1. **Native Avro Viewer Support**  
   `<SchemaViewer />` is extended to parse and render Avro types directly (including unions, logical types, named references, etc.).

**or**

2. **Automatic Avro → JSON Schema Conversion (on the fly or at build time)**  
   When a `.avsc` file is passed to `<SchemaViewer />`, EventCatalog internally converts it into JSON Schema and then renders it via the existing viewer component.

Either approach would allow Avro users to benefit from the richer schema navigation UI without changing their schema tooling or duplicating schema definitions.

### Implementation Notes

- The maintainer mentioned that **`avro-js`** can parse `.avsc` schemas in JavaScript, which could help form the basis of either:
  - A direct viewer renderer, or
  - A transformation layer that outputs structured JSON compatible with `<SchemaViewer />`.

- Tools that may assist in Avro → JSON Schema transformation (not necessarily adoption recommendations, but references):
  - https://github.com/clemensv/avrotize
  - https://github.com/jhasselgren/avro-to-json-schema
  - https://github.com/fge/json-schema-avro

- Happy to provide example schemas and help test early prototypes.

Thanks again for the great work on EventCatalog — love to help move this forward! 🎉

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
[...]

PR #1820, opened 2025-11-11
fix(schema-viewer): display required flags for Avro schemas

mumundum:
### Summary
While testing `@eventcatalog/core@2.64.1`, I noticed that the new `showRequired` prop was not being applied:
- Avro required field badges were never shown because `showRequired` was not passed down to `AvroSchemaViewer` from its parent render contexts.

### Context
Discussed originally in #1799.  
Verified with a clean install of 2.64.1 and an Avro schema using `<SchemaViewer file="schema.avro" showRequired="true" />`.

### DIFF
--- a/.changeset/cold-toys-swim.md
+++ b/.changeset/cold-toys-swim.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+fix(core): Schemaviewer required flags for Avro schemas now working


boyney123:
Thank you for the fix @mumundum 🙌

boyney123:
@all-contributors add @mumundum for code

PR #1947, opened 2025-12-24
Fix incorrect version-specific channel routing in NodeGraph visualiser

mjftw:
### Context

I was using EventCatalog on my own catalogue and wanted to add an example to show how a service might dual publish the same event on two different major versions, to two different channels (different Kafka topics in this case).

I found that the catalogue was not rendering how I had expected it to. I thought it was an EventCatalog bug, so I decided to see if I could reproduce it in the examples and fix it rather than putting up a Github Issue and hoping someone else would fix it.

### Issue as seen in my Event Catalogue

Here's what it looked like in my service's .mdx frontmatter:

```yaml
---
id: example_orders_service
name: Example Orders Service
version: 1.0.0
sends:
  - id: event:example.order.created
    version: 1.x
    to:
      - id: channel:example.order.created
        version: 1.x
  - id: event:example.order.created
    version: 2.x
    to:
      - id: channel:example.order.created
        version: 2.x
---
```

And in the UI:

Screenshot 2025-12-24 at 16 13 30 (1)
Screenshot 2025-12-24 at 16 21 44 (1)

In my Service frontmatter version 1.x of the event should go to 1.x of the channel, and 2.x event to 2.x channel, but that's not what the visualiser shows, so I think this is a bug.

This PR fixes incorrect event-to-channel routing in NodeGraph visualisation when a service sends multiple versions of the same event to different version-specific channels.

Here is a screenshot taken from my catalogue after using the fixed version of `@eventcatalog/core` (changes in this PR), showing the issue resolved.

Screenshot 2025-12-30 at 09 45 38

More info on the bug and fix below.

### The Problem - reproduced in EventCatalog source tree

I was able to reproduce the issue in the EventCatalog examples by setting up a similar scenario.

When a service is configured to send different versions of an event to different channels:

```yaml
sends:
  - id: InventoryAdjusted
    version: 0.0.1
    to:
      - id: inventory-adjusted-channel
        version: 0.0.1
  - id: InventoryAdjusted
    version: 1.0.1
    to:
      - id: inventory-adjusted-channel
        version: 1.0.1
```

The NodeGraph visualisation would incorrectly show **both** event versions routing to the **same** channel (whichever was found first), rather than each version routing to its correct version-specific channel.

### Screenshots

**Before (Bug):** InventoryAdjusted v1.0.1 incorrectly routes to `inventory-adjusted-channel` v0.0.1

Screenshot 2025-12-24 at 17 29 18

**After (Fix):** InventoryAdjusted v1.0.1 correctly routes to `inventory-adjusted-channel` v1.0.1

Screenshot 2025-12-24 at 17 57 08

### Root Cause

The `.find()` method in `message-node-graph.ts` and `services-node-graph.ts` was only matching by `id`, ignoring the `version` field:

```typescript
// Before: Only matched on id, ignoring version
const config = sends.find((send) => send.id === message.data.id);
```

This caused the first matching entry to be returned regardless of version.

### The Fix

1. Added a generic `versionMatches()` utility function to `utils/collections/util.ts` that correctly handles version matching with comprehensive pattern support:
   - **Exact versions** (1.0.0 matches 1.0.0)
   - **Semver ranges** (^1.0.0, ~1.2.0, >=1.0.0 <2.0.0) using the `semver` package
   - **X-patterns** (1.x, 1.2.x) with correct boundary checks
   - **`'latest'` keyword** (matches any version)
   - **`undefined` version** (matches any version)

2. Updated the node graph specific `versionMatches()` helper in `utils/node-graphs/utils/utils.ts` to use the generic utility whilst handling graph-specific edge cases (`undefined` and `'latest'`)

3. Updated `getProducersOfMessage()` and `getConsumersOfMessage()` in `utils/collections/services.ts` to use the new `versionMatches()` utility, replacing direct `semver.satisfies()` calls

4. Updated all relevant `.find()` calls in `message-node-graph.ts` (6 locations) and `services-node-graph.ts` (2 locations) to include version matching:

```typescript
// After: Matches on both id AND version, with semver/x-pattern support
const config = sends.find(
  (send) => send.id === message.data.id && versionMatches(send.version, message.data.version)
);
```

## Semver and X-Pattern Support

The implementation now supports semver ranges and x-pattern matching!

Services can now use flexible version patterns in their `sends` and `receives` configurations:

```yaml
# Caret ranges - accepts compatible versions
sends:
  - id: OrderCreated
    version: ^1.0.0  # Matches 1.0.0, 1.2.3, 1.9.9, but not 2.0.0

# Tilde ranges - accepts patch-level changes
receives:
  - id: PaymentProcessed
    version: ~1.2.0  # Matches 1.2.0, 1.2.5, but not 1.3.0

# X-patterns - wildcard matching
receives:
  - id: InventoryAdjusted
    version: 1.x     # Matches any 1.x.x version
  - id: OrderCreated
    version: 1.2.x   # Matches 1.2.0, 1.2.5, but not 1.3.0
```

This enables more flexible version management strategies while maintaining correct graph visualisation and routing.

When a pattern matches multiple catalogue versions, the latest matching version is resolved — keeping the graph clean and unambiguous.

## Example Catalogue Demonstrates Version Patterns

The example catalogue now includes a minimal demonstration of the three version matching patterns:

Run the catalogue:
```bash
pnpm run start:catalog
```

Navigate to the service visualisers to see version pattern matching in action:
- **http://localhost:3000/visualiser/services/InventoryService/0.0.2** - Sends `InventoryAdjusted` v1.0.0 (exact match)
Screenshot 2026-02-20 at 18 52 27

- **http://localhost:3000/visualiser/services/NotificationService/0.0.2** - Receives `InventoryAdjusted` ^1.0.0 (resolves to latest match)
Screenshot 2026-02-20 at 18 53 08

- **http://localhost:3000/visualiser/services/OrdersService/0.0.3** - Receives `InventoryAdjusted` 1.x (resolves to latest match)
Screenshot 2026-02-20 at 18 53 32

### Unit Tests
- **`eventcatalog/src/utils/__tests__/node-graphs/utils/utils.spec.ts`** - Tests for graph-specific `versionMatches()` wrapper including delegation to generic utility
- **`eventcatalog/src/utils/__tests__/collections/util.spec.ts`** - Comprehensive tests for generic `versionMatches()` utility covering exact versions, semver ranges (^, ~, complex), x-patterns, latest keyword, and boundary edge cases

### Integration Tests
- **`eventcatalog/src/utils/__tests__/messages/node-graph.spec.ts`** - Integration tests verifying:
  - Version-specific channel routing (original bug fix)
  - Producer/consumer matching with caret ranges (^1.0.0)
  - Tilde range matching (~1.2.0)
  - X-pattern matching at major (1.x) and minor (1.2.x) levels
  - Semver patterns with channel configurations
  - Explicit multi-version receives (service listing multiple specific versions)

### DIFF
--- a/packages/core/eventcatalog/src/utils/__tests__/messages/catalog/channels/EventBridgeChannel/index.mdx
+++ b/packages/core/eventcatalog/src/utils/__tests__/messages/catalog/channels/EventBridgeChannel/index.mdx
@@ -1,0 +1,9 @@
+---
+id: EventBridgeChannel
+name: EventBridge Channel
+version: 1.0.0
+routes:
+  - id: SQSChannel
+    version: 1.0.0
+---
+## EventBridge Channel
--- a/packages/core/eventcatalog/src/utils/__tests__/messages/catalog/channels/SNSChannel/index.mdx
+++ b/packages/core/eventcatalog/src/utils/__tests__/messages/catalog/channels/SNSChannel/index.mdx
@@ -1,0 +1,6 @@
+---
+id: SNSChannel
+name: SNS Channel
+version: 1.0.0
+---
+## SNS Channel
--- a/packages/core/eventcatalog/src/utils/__tests__/messages/catalog/channels/SQSChannel/index.mdx
+++ b/packages/core/eventcatalog/src/utils/__tests__/messages/catalog/channels/SQSChannel/index.mdx
@@ -1,0 +1,9 @@
+---
+id: SQSChannel
+name: SQS Channel
+version: 1.0.0
+routes:
+  - id: SNSChannel
+    version: 1.0.0
+---
+## SQS Channel
--- a/packages/core/eventcatalog/src/utils/__tests__/messages/catalog/events/PaymentProcessed/index.mdx
+++ b/packages/core/eventcatalog/src/utils/__tests__/messages/catalog/events/PaymentProcessed/index.mdx
@@ -1,0 +1,6 @@
+---
+id: PaymentProcessed
+name: Payment Processed
+version: 0.0.1
+---
+## Payment Processed
--- a/packages/core/eventcatalog/src/utils/__tests__/messages/catalog/services/OrderService/index.mdx
+++ b/packages/core/eventcatalog/src/utils/__tests__/messages/catalog/services/OrderService/index.mdx
@@ -1,0 +1,12 @@
+---
+id: OrderService
+name: Order Service
+version: 1.0.0
+receives:
+  - id: PaymentProcessed
+    version: 0.0.1
+    from:
+      - id: SNSChannel
+        version: 1.0.0
+---
+## Order Service
--- a/packages/core/eventcatalog/src/utils/__tests__/messages/catalog/services/PaymentService/index.mdx
+++ b/packages/core/eventcatalog/src/utils/__tests__/messages/catalog/services/PaymentService/index.mdx
@@ -1,0 +1,12 @@
+---
+id: PaymentService
+name: Payment Service
+version: 1.0.0
+sends:
+  - id: PaymentProcessed
+    version: 0.0.1
+    to:
+      - id: EventBridgeChannel
+        version: 1.0.0
+---
+## Payment Service


mjftw:
@boyney123 - any thoughts? 🙏🏽

boyney123:
Thanks for finding this @mjftw and writing the details response helped me understand. Looks like a bug for sure.

Your questions

> Currently the changes will only allow matching on the exact version if specified (e.g. 1.0.0 matches 1.0.0) - would it be better to also allow semver and x matching? (e.g. ^1.0.0 or 1.x)

Yeah I think we should, we match on semver and x matching for other references when we use versions I believe, what do you think? could be a good addition, are you OK to add that?

Also think some minor conflicts in your PR, sorry it took me so long to get back on this.

mjftw:
> Thanks for finding this @mjftw and writing the details response helped me understand. Looks like a bug for sure.
> 
> Your questions
> 
> > Currently the changes will only allow matching on the exact version if specified (e.g. 1.0.0 matches 1.0.0) - would it be better to also allow semver and x matching? (e.g. ^1.0.0 or 1.x)
> 
> Yeah I think we should, we match on semver and x matching for other references when we use versions I believe, what do you think? could be a good addition, are you OK to add that?
> 
> Also think some minor conflicts in your PR, sorry it took me so long to get back on this.

Hello, no problem, yep will do!

mjftw:
@boyney123 - As per our discussion I've updated the PR to resolve matches based on:

1. Exact version match (e.g. 1.0.0)
2. Semver range matching (e.g. ^1.0.0)
3. N.x matching (e.g. 1.x)

A consequence of this is that I needed to make it possible for multiple nodes to be found in a match.

E.g. a Reference with pattern ^1.0.0 matches versions 1.0.1 & 1.2.3, and it should therefore show in the graph that it matches those, but previously it would only show one of these, which feels like misleading/incorrect behaviour to me.

Check this out in the screenshots I added to the PR in the manual testing section to see a *before* and *after* the changes (both images were generated including the the new catalog mdx file changes I made to demo the matching).

Apologies, with these changes the PR got a little larger and more complex than originally intended. I changed from a _patch_ version bump to a _minor_ bump as I think this adds new functionality (multiple matches), but feel free to change it back to a _patch_ only if that better matches how you've been doing your versioning.

Thanks!

mjftw:
Actually I've been thinking through the multiple matches per pattern thing and I'm not sure it's correct after all.

If a service declares it publishes version ~1.0.0 of an event and versions 1.0.1 and 1.2.3 exist, it probably actually publishes only the newest matching version (1.2.3) and is unlikley to dual punlish 1.0.1 and 1.2.3.
However I think the expectation is reveresed on the consuming side?
E.g. If a service declares it consimes version 1.x of an event and versions 1.0.1 and 1.2.3 exist, it probably actually does consume both 1.0.1 and 1.2.3 (e.g. it may have a wildcard match in its event routing and is built to handle both versions).

Having this difference on producer and consumer side gets complex and confusing though.

One option would be to allow the relationship to specify explicitly.
E.g. alongside specifying the version range you can also specify whether that is inclusive of all matches or just the latest.
```
sends:
  - id: MyEvent
    version: 1.x
    match: all OR latest
```

And we'd pick sane defaults for whether 'all' or 'latest' is chosen for sends & recieves.

This again gets kind of complicated though.


What do you think should be the correct behaviour?

boyney123:
`If a service declares it publishes version ~1.0.0 of an event and versions 1.0.1 and 1.2.3 exist, it probably actually publishes only the newest matching version (1.2.3) and is unlikley to dual punlish 1.0.1 and 1.2.3.`

Yeah I think this is right, we should default to the latest one here.

TBH I think the same for consumers (which I thought already was the case), but lets match on the latest using that semver pattern matching. Then we dont infer anything or guess the intentions.

I think this keeps it clear/clean. You can use exact matching like you can... but if you use semver matching it will find the latest one using that match (if that makes sense).

I think if people want to publish multiple messages  (differnet versions) they should be expliect and just put down that it raises those two messages with version (whatever).

WYT?

mjftw:
@boyney123 I've now updated this PR to drop the functionality added around getting multiple matches as suggested (this is much simpler too), and I've rebased on main.
I've confirmed everything is working, but I'd suggest you checkout the branch and run the manual testing instructions yourself just to be sure 🙂 

This PR is now hopefully good to go 🤞🏽

boyney123:
Great thanks for the effort here @mjftw !

Just a quick one try run `npm run format` which should format the code for the linter.

mjftw:
> Great thanks for the effort here @mjftw !
> 
> Just a quick one try run `npm run format` which should format the code for the linter.

No problem, done!

mjftw:
Ack reading through the PR I think some things got messed up during the rebase following the move to monorepo setup.
Will fix.

mjftw:
@boyney123 - Okay I think we're back to just changes that belong on this branch. 🤞🏽

boyney123:
Thanks @mjftw I'm happy for us to get this in and move forward with it :), any side effects or small issues we might find, we can roll foward and fix. But looking at it all makes sense to me!

Thanks so much for the contribution!

Issue #2067, opened 2026-02-04
Add typed references to resolve namespace collisions in metadata


mjftw:
### Use Case

When building an EventCatalog, we often encounter namespace collisions where resources of different types share the same ID. This creates ambiguity that prevents the catalog from correctly resolving references.

**Concrete examples:**

1. **Owners field**: A catalog might have both a user `john-doe` and a team `john-doe`. When you write `owners: ['john-doe']`, EventCatalog cannot determine whether to link to the user or the team.

2. **Sends/receives fields**: Multiple message types can share IDs. For instance, `OrderCreated` could be an event, command, or query. When you write `sends: [{ id: 'OrderCreated' }]`, the catalog cannot determine which type to link to, so it will pick one, which may not be what was intended.

3. **Flow step messages**: The same ambiguity occurs in flow diagrams where message steps need to reference specific message types but can't distinguish between events, commands, and queries with the same ID.

This problem becomes particularly acute in larger catalogs where:
- Naming conventions lead to ID reuse across different resource types
- Teams work independently and create overlapping identifiers
- Documentation clarity requires precise linking to specific resource types

Currently, users have no way to disambiguate these references, leading to incorrect links, broken references, or the need to artificially rename resources to avoid collisions.

### Proposed Solution

Introduce **wiki-style typed references** using the syntax `type|id` across all metadata reference fields, whilst maintaining full backward compatibility with existing plain string/object references.

We can match the reference syntax used here:
https://www.eventcatalog.dev/docs/development/components/resource-references

**User experience:**

Instead of writing ambiguous references:
```yaml
owners: ['john-doe', 'platform-team']
```

Users could explicitly specify types:
```yaml
owners: ['user|john-doe', 'team|platform-team']
```

For message references:
```yaml
sends: 
  - id: 'event|OrderCreated'
    version: '1.0.0'
  - id: 'command|ProcessOrder'
```

**Expected behaviour:**

1. When a typed reference like `user|john-doe` is used, EventCatalog should:
   - Only look up the resource in the specified collection (users)
   - Create links specifically to that resource type

2. When a plain reference like `john-doe` is used (existing behaviour):
   - Continue checking all possible types (backward compatible)
   - Optionally warn if multiple matches are found (collision detection)

3. The syntax should work across all reference fields:
   - `owners` (user | team)
   - `sends`/`receives` (event | command | query)
   - Flow step messages (event | command | query)
   - Any future reference fields

**Benefits to users:**

[...]

PR #2075, opened 2026-02-05
Type scoped entity references

mjftw:
## Motivation

See issue: https://github.com/event-catalog/eventcatalog/issues/2067


boyney123:
Hey @mjftw is this still a core issue for you? Just cleaning up GH at the moment, and trying to figure out what we should re look at / keep for now.

mjftw:
> Hey @mjftw is this still a core issue for you? Just cleaning up GH at the moment, and trying to figure out what we should re look at / keep for now.

Hi @boyney123, to be honest as much as it would be useful I think it's more a nice to have than a key feature for us now. I thought this would cause issues for us but far it's not been a problem, so nothing too urgent.

Closing the PR.

PR #2082, opened 2026-02-10
fix(core): resolve PlantUML rendering race condition with destroyZoom…

ZakaryaCH:
…Instances

## Motivation

  PlantUML diagrams fail to render on pages that have no Mermaid diagrams (e.g., a domain index page). The placeholder
  text "Rendering PlantUML diagram..." stays visible indefinitely with no errors in the console.

  Root cause: Race condition between destroyZoomInstances() and renderPlantUMLWithZoom() in mermaid-zoom.ts. When
  astro:page-load fires, the Mermaid listener calls destroyZoomInstances() which sets renderingAborted = true. Since
  there are no Mermaid diagrams on the page, renderMermaidWithZoom() is never called and renderingAborted stays true.
  The PlantUML listener resets the flag, but the subsequent await import('pako') creates an async yield during which the
   flag can be set back to true, silently aborting the PlantUML rendering.

  Fix: Move renderingAborted = false to after await import('pako'), right before the rendering loop, so the reset
  happens after all async operations that could be interrupted by destroyZoomInstances().

  Reproduction:
  1. Create a page with a PlantUML diagram and no Mermaid diagrams
  2. Navigate to the page — PlantUML shows "Rendering PlantUML diagram..." forever
  3. Calling renderPlantUML(document.getElementsByClassName('plantuml')) manually in the browser console renders the
  diagram successfully

Tested locally — PlantUML diagrams render correctly on page load after the change.

### DIFF
--- a/.changeset/slimy-planes-exist.md
+++ b/.changeset/slimy-planes-exist.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+fix(core): resolve PlantUML rendering race condition


boyney123:
Thanks for finding this and the fix @ZakaryaCH

boyney123:
@all-contributors add @ZakaryaCH for bug

PR #2094, opened 2026-02-10
Remove 'max-h-content' class from content

jboeijenga:
## Motivation

When embedding EventCatalog content, the height of the application is not calculated correctly.
Because of this, the bottom part of the application gets cut off in embed mode.

Screenshot 2026-02-10 at 17 02 51

The `max-h-content` class is applied to the content element in embed mode.
This class sets the height to: `calc(100vh - 4rem);`
This is presumably intended to reserve space for the 4rem header in normal mode.
However, in embed mode no header is rendered, meaning the height is unnecessarily reduced. 
As a result, the bottom of the application gets cut off.

Removing the `max-h-content` class in embed mode allows the application to use the full available height and prevents the bottom content from being clipped.

### DIFF
--- a/.changeset/chilly-shoes-hug.md
+++ b/.changeset/chilly-shoes-hug.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+chore(core): removed 'max-h-content' class from content


boyney123:
Thanks @jboeijenga makes sense to me, thanks for raising this! 

If you could review the CLA then we can get it merged :)

jboeijenga:
@boyney123: thanks for the quick review, I just agreed with the CLA.

boyney123:
Thanks @jboeijenga

boyney123:
@all-contributors add @jboeijenga for code

PR #2136, opened 2026-02-15
perf: disable full catalog API by default and gate Studio copy flow

thorthelegend99:
## Summary
- set `api.fullCatalogAPIEnabled` default to `false` for `/api/catalog`
- update config typing/docs to reflect the new default
- when disabled, Studio CTA opens EventCatalog Studio directly without the copy-resources flow

## Why
On large catalogs, pre-rendering `/api/catalog` is very expensive. In local perf tests (~1027 resources), disabling this endpoint reduced build time from ~102s to ~55s (about 46% faster).

## Notes
- The full catalog API can still be enabled explicitly via:
  ```js
  ```
- Formatter run: `corepack pnpm --filter @eventcatalog/core format`


thorthelegend99:
Superseded by #2137, which is rebased on current main and contains the same intended changes.

PR #2137, opened 2026-02-15
perf(core): default full catalog API off + dedupe shared loader work

thorthelegend99:
## Summary
- set `api.fullCatalogAPIEnabled` default to `false` for `/api/catalog`
- update config typing/docs to reflect the new default
- when disabled, Studio CTA opens EventCatalog Studio directly without the copy-resources flow
- add in-flight promise dedupe for shared loaders used across many routes (`getDomains`, `getServices`, `getNestedSideBarData`) to avoid duplicate concurrent work during build

## Why
For large catalogs, full static builds were spending a lot of time redoing shared work.

On local perf tests (~1027 resources):
- `/api/catalog` on: ~102s
- `/api/catalog` off by default: ~55s
- with in-flight memoization: ~57.5s in current config path (keeps behavior stable while preventing concurrent duplicate loader work)

## Notes
- Full catalog API can still be enabled explicitly:
  ```js
  ```
- Formatter run: `corepack pnpm --filter @eventcatalog/core format`
- This PR is based on current `main` and supersedes the earlier branch-based PR.

### DIFF
--- a/packages/core/eventcatalog/src/stores/sidebar-store/state.ts
+++ b/packages/core/eventcatalog/src/stores/sidebar-store/state.ts
@@ -24,6 +24,7 @@ export type { NavigationData, NavNode, ChildRef };
 
 const CACHE_ENABLED = process.env.DISABLE_EVENTCATALOG_CACHE !== 'true';
 let memoryCache: NavigationData | null = null;
+let inFlightCache: Promise<NavigationData> | null = null;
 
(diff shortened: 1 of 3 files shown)


thorthelegend99:
Split into separate PRs per request: #2138 (API default/studio gating) and #2139 (in-flight loader dedupe).

PR #2138, opened 2026-02-15
perf(core): disable full catalog API by default

thorthelegend99:
## Summary
- change `/api/catalog` default to disabled (`api.fullCatalogAPIEnabled ?? false`)
- update config typing and API docs to reflect default false
- when disabled, Studio CTA opens EventCatalog Studio directly

## Why
For large catalogs, prerendering `/api/catalog` adds substantial build time and memory pressure.

## Enable if needed
```js
```

## Validation
- formatter run: `corepack pnpm --filter @eventcatalog/core format`
- benchmarked on ~1k resource synthetic catalog (shared separately)

### DIFF
--- a/.changeset/olive-pumas-destroy.md
+++ b/.changeset/olive-pumas-destroy.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+perf(core): disable full catalog API by default


thorthelegend99:
Closing as superseded by #2149 for route/integration behavior. We can follow up with a small UI-only PR for any remaining Studio gating if needed.

PR #2139, opened 2026-02-15
perf(core): dedupe in-flight shared loader work

thorthelegend99:
## Summary
- add in-flight promise dedupe to shared high-traffic loaders so concurrent callers reuse the same work during builds:
  - `getDomains`
  - `getServices`
  - `getNestedSideBarData`

## Why
Static route generation can trigger overlapping calls to shared loaders. Existing memory caches help after completion, but not while a load is still in progress. This change prevents duplicate concurrent computation.

## Benchmark note
On the current ~1k resource synthetic benchmark path with full catalog API disabled, this was roughly neutral in wall-clock time (run-to-run variance dominates):
- without in-flight dedupe: ~56.99s
- with in-flight dedupe: ~57.54s

This suggests the change is safe, but likely not a major win in this scenario.


boyney123:
@codex review

boyney123:
@codex review

PR #2147, opened 2026-02-16
perf(search): compact API payload for search index

thorthelegend99:
## Summary
Reduces search index transfer size by compacting API payload keys while keeping frontend backward-compatible.

### Changes
- `GET /api/search-index.json` now returns compact shape:
  - top-level: `{ i: [...] }` instead of `{ items: [...] }`
  - per item: `{ k, t, b?, s?, h? }` instead of `{ key, title, badge?, summary?, href? }`
- Omits optional fields when empty to reduce payload bytes further.
- Adds `Vary: Accept-Encoding` header to explicitly support compression-aware caches/CDNs.
- Search modal now supports both payload formats:
  - new compact shape (`i`)
  - previous shape (`items`) for compatibility.

## Why
Search index loading is now on-intent; shrinking that payload gives a direct first-open latency win on larger catalogs.

## Validation
- formatter run: `corepack pnpm --filter @eventcatalog/core format`
- sanity test run: `corepack pnpm --filter @eventcatalog/core exec vitest run eventcatalog/src/utils/__tests__/url-builder.spec.ts`

### DIFF
--- a/.changeset/serious-years-sin.md
+++ b/.changeset/serious-years-sin.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+perf(search): compact API payload for search index


boyney123:
@codex review

PR #2162, opened 2026-02-16
Add in-memory file index cache to SDK for faster catalog lookups

boyney123:
## What This PR Does

Adds an automatic in-memory file index cache to the SDK that eliminates repeated filesystem scans. Previously, every `findFileById`, `versionExists`, and `getResource` call triggered a full recursive glob + file reads across the entire catalog directory. Now a single scan builds an in-memory index on first access, and all subsequent lookups are O(1) map gets.

Also fixes the playground URL in the CLI export command from `localhost:5173` to `https://playground.eventcatalog.dev/`.

### Key Changes
- **In-memory file index cache** (`internal/utils.ts`): Auto-built on first read via `ensureFileCache()`. Stores file paths, IDs, versions, and parsed frontmatter in Maps for instant lookups.
- **Cache invalidation**: Every SDK write/rm/version operation across all 13 resource files calls `invalidateFileCache()` to keep the cache consistent.
- **Stale cache detection**: Checks catalog directory mtime to detect external changes (e.g., test `beforeEach` wipes).
- **Message type index** (`dsl/utils.ts`): Replaces per-message recursive glob calls with a single `buildMessageTypeIndex()` that does 3 globs once, then O(1) Map lookups. Threaded through service/domain DSL conversion via `_msgIndex` option.
- **Cached matter.read** (`cachedMatterRead`): Returns pre-parsed frontmatter from cache when available, avoiding redundant file reads during hydration.
- **Playground URL fix** (`cli/export.ts`): Changed from `http://localhost:5173/` to `https://playground.eventcatalog.dev/`.

## How It Works

The cache is completely automatic - no opt-in required:

1. First `findFileById` call triggers `buildFileCache()`: one glob for all `index.{md,mdx}` files, reads each once with `gray-matter`, stores results in two Maps (`_fileIndexCache` by ID, `_matterCache` by path)
2. Subsequent reads hit the Maps directly - O(1) instead of O(n) glob + file reads
3. Any SDK mutation (`writeResource`, `rmResourceById`, `versionResource`, `rmEvent`, `writeTeam`, etc.) calls `invalidateFileCache()` which nulls the Maps
4. Next read after a write rebuilds the cache automatically

For DSL export with hydration, this reduces ~100+ recursive filesystem scans to just 1.

## Test plan
- [x] All 574 SDK tests pass
- [x] All 54 CLI tests pass
- [ ] Manual test: `npx @eventcatalog/cli export --resource service OrdersService --hydrate` completes in <2s vs ~60s before

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/sdk/src/dsl/utils.ts
+++ b/packages/sdk/src/dsl/utils.ts
@@ -73,10 +73,11 @@ export function buildMessageTypeIndex(catalogDir: string): MessageTypeIndex {
   for (const type of types) {
     const matches = globSync(`**/${type}/*/index.{md,mdx}`, { cwd: catalogDir });
     for (const match of matches) {
-      const parts = match.split('/');
+      const parts = match.replace(/\\/g, '/').split('/');
       const typeIdx = parts.lastIndexOf(type);
       if (typeIdx !== -1 && typeIdx + 1 < parts.length) {
-        index.set(parts[typeIdx + 1], typeMap[type]);
+        const id = parts[typeIdx + 1];
+        if (!index.has(id)) index.set(id, typeMap[type]);
       }
     }
   }


thorthelegend99:
Great performance direction here — this is a strong win overall. I spotted two important issues to fix before merge:

1) **Message-type precedence regression** (`packages/sdk/src/dsl/utils.ts`)
   - `buildMessageTypeIndex()` currently overwrites existing IDs as it loops `events -> commands -> queries`.
   - That changes behavior vs previous `resolveMessageType()` fallback, which effectively had precedence `event > command > query` (first match wins).
   - With current map writes, precedence becomes `query > command > event` when IDs collide.
   - **Suggested fix:** preserve first match, e.g. `if (!index.has(id)) index.set(id, typeMap[type]);`

2) **Cross-platform path parsing bug** (`packages/sdk/src/dsl/utils.ts`)
   - `const parts = match.split('/')` can fail on Windows where separators can be `\`.
   - This can break message ID extraction in the index.
   - **Suggested fix:** normalize first, e.g. `const parts = match.replace(/\\/g, '/').split('/')` (or equivalent path-safe parsing).

Non-blocking: catalog-dir `mtime` is a weak stale-cache signal for nested edits, but explicit invalidation on SDK writes covers most normal flows.

PR #2183, opened 2026-02-18
Feature/optimistic cache

ababilone:
## Motivation

SDK operations on big catalog keep getting slower & slower the more mdx files there is in it.

Using the recently introduced caching mechanism, this PR is an attempt to change its frequent invalidation to use instead optimistic changes related to the currently applied alteration.

First test on a big catalog show great performance improvement (down to ~5ms writeResource operations coming from up to 1500-2000ms/operation)

### DIFF
--- a/.changeset/metal-fans-tan.md
+++ b/.changeset/metal-fans-tan.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/sdk": minor
+---
+
+Implement optimistic cache management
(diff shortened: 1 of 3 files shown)


boyney123:
Thanks for this @ababilone I was working on a PR (work in progresS) https://github.com/event-catalog/eventcatalog/pull/2198

That actually picked up some performance improvements, I'm going to get them out, and see how you get on if thats OK.

I think they are very simular

boyney123:
going to close this for now in favour of the https://github.com/event-catalog/eventcatalog/pull/2198 that was done, let's review again if we still think this is needed, thanks @ababilone !

PR #2288, opened 2026-03-04
fix(core): allow multiple openapi specs per service in schema explorer

ugostack:
Without this fix, services with multiple openapi specs shared the same group key, causing the schema explorer to mismatch specs between them.

## Motivation

(Write your motivation here.)

### DIFF
--- a/.changeset/strong-facts-tie.md
+++ b/.changeset/strong-facts-tie.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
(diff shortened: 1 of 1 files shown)


boyney123:
Thanks for this @ugostack , happy to get this in, I wanted to make a couple of edits to the PR but was struggling my side for whatever reason, I made a new PR to fix this, so will close this one (pretty much the same thing). 

Thanks!

boyney123:
This is now out in the latest version of EventCatalog :)

boyney123:
@all-contributors add @ugostack for code

Issue #2303, opened 2026-03-05
addEventToService removes all events and commands files


callingmybluff:
### I tried this:

Executed the following in order

1. `writeEventToService`
2. `addEventToService`

### This happened:

The event file is removed (by the 2nd sdk call), and is listed under the service.

### I expected this:

The event file to exist

### Is there a workaround?

Yes, change the order!

However, if you have multiple additions / writes, you need to do all your additions first, then all your writes.

### Anything else?

Link to discord bug https://discord.com/channels/918092420338569216/1479187694017056768

### EventCatalog Version

@eventcatalog/sdk@{2.17.0}

### Node.js Version

v22.15.0

### Platform(s)

Linux

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.

PR #2304, opened 2026-03-05
fix: #2303 removing service to rewrite it persist existing files

callingmybluff:
## Motivation

I was working on adding events and commands automatically, and this was blocking my progress.


boyney123:
Have some problems with CI/CD going to close this and reopen with your fix.



Thanks for the fix @callingmybluff

PR #2567, opened 2026-05-24
fix(sdk): use real RegExp in addMessageToChannel path split

jagreehal:
## Summary

`addMessageToChannel` (and its `addEventToChannel` / `addCommandToChannel` / `addQueryToChannel` bindings) was rewriting messages into a duplicate nested path, turning `index.mdx` into a directory:

```
catalog/events/OrderPlaced/index.mdx/events/OrderPlaced/index.mdx
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       a directory, not a file
```

The cause is one line in `packages/sdk/src/channels.ts:278`:

```ts
const path = existingResource.split(`/[\\/]+${collection}`)[0];
```

That template *string* — `` `/[\\/]+${collection}` `` — is passed as a literal substring to `String.prototype.split`, not as a `RegExp`. The literal never matches real paths, so `.split()` returns the whole input unchanged. The downstream `join(path, collection)` then nests a duplicate copy of the resource under `<orig>/<collection>/<id>/index.mdx`, and the subsequent `writeMessage` creates `index.mdx` as a directory.

The equivalent calls in `services.ts` already use a real `RegExp`:

```ts
// services.ts (correct)
const path = existingResource.split(/[\\/]+services/)[0];
```

This PR makes `channels.ts` consistent.

## Changes

- **`packages/sdk/src/channels.ts`**: replace template-string split with `new RegExp('[\\\\/]+' + collection)` so it matches both POSIX and Windows separators.
- **`packages/sdk/src/test/channels.test.ts`**: add a focused regression test under the existing `addEventToChannel` describe. It writes a channel + event, calls `addEventToChannel`, then asserts (a) the canonical `events/<id>/index.mdx` is still a *file*, and (b) the bad nested path `events/<id>/index.mdx/events/<id>/index.mdx` does not exist. Reverting the channels.ts change makes this test fail with `expected false to be true` on `indexMdxStat.isFile()`, confirming it catches the regression.
- **`.changeset/fix-addmessagetochannel-regex-split.md`**: patch changeset for `@eventcatalog/sdk`.

## How discovered

While building a consumer (`autotel-eventcatalog generate`, which uses `@eventcatalog/sdk` to scaffold catalog resources from runtime telemetry snapshots), my generated event files ended up nested as described above. Tracing through the SDK build pinpointed the string-vs-regex line.

## Test plan

- [x] `pnpm test` in `packages/sdk` — 26 files, **733 / 733 tests pass** including the new regression.
- [x] Revert just the `channels.ts` change locally and re-run: regression test fails as expected, confirming it catches the bug.
- [x] No behaviour change for happy-path callers — existing tests in `channels.test.ts` (including the three pre-existing `addEventToChannel` cases) pass unchanged.


boyney123:
Thanks for this @jagreehal , are you OK to read/sign https://cla-assistant.io/event-catalog/eventcatalog?pullRequest=2567, do we can get it merged?

Thanks!

jagreehal:
Thanks, @boyney123 Just signed the CLA.

boyney123:
Having various issues with lock files, linting etc... going to reopen this PR, thanks for fixing it, appreciate the time!

boyney123:
https://github.com/event-catalog/eventcatalog/pull/2584

PR #2570, opened 2026-05-24
chore(deps): use fetch instead of axios

jdebarochez:
Use native fetch instead of axios in analytics. I ran `pnpm install` afterward but I'm not confident in its output if the reviewer can pay attention to every line changed. I would expect only a removal of axios reference.

## Motivation

Axios library is only used in one place and can be replaced with native fetch library. It's already the case in `packages/create-eventcatalog/templates/analytics.ts`. This project can remove its dependency from axios and would have one less library to maintain. 

A better rationale is available on Node.js blog at https://nodejs.org/en/blog/migrations/axios-to-fetch

### DIFF
--- a/packages/core/src/analytics/analytics.js
+++ b/packages/core/src/analytics/analytics.js
@@ -1,4 +1,3 @@
-import axios from 'axios';
 import os from 'os';
 import { VERSION } from '../constants';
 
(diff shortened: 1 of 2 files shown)


boyney123:
Thanks @jdebarochez for this, makes perfect sense. 

Not sure what happened with lock file and conflicts, I made another PR to fix that https://github.com/event-catalog/eventcatalog/pull/2573, so will close this one, but thanks for this contribution I appreciate it.

boyney123:
@all-contributors add @jdebarochez for code

PR #2604, opened 2026-06-04
fix(core): remove unused concurrently

Enteris:
## Summary
This PR removes concurrently from the core package dependencies because it is no longer used.

## Why
Core previously used concurrently for local dev process orchestration, but according to [release 3.15.3](https://github.com/event-catalog/eventcatalog/releases/tag/%40eventcatalog%2Fcore%403.15.3), it was replaced with native process spawning. Additionally, version 8 of concurrently have some transitive vulnerabilities.

Validated that the code was no longer used, and removed the dependency and updated the lockfile.

## What changed
- Removed concurrently from the core package dependency list.
- Updated lockfile accordingly.

### DIFF
--- a/.changeset/green-chefs-wave.md
+++ b/.changeset/green-chefs-wave.md
@@ -1,0 +1,7 @@
+---
+'@eventcatalog/core': patch
+---
+
+Remove the unused `concurrently` dependency from `@eventcatalog/core`.
+Usage was removed in [3.15.3](https://github.com/event-catalog/eventcatalog/releases/tag/%40eventcatalog%2Fcore%403.15.3) 
+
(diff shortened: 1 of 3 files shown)


boyney123:
thanks for this @Enteris I had to clean up some other things too so I merged this into another PR https://github.com/event-catalog/eventcatalog/pull/2612

Thanks for the PR and your time!

Enteris:
I was toying with knip and wondered if you wanted to add it to the project, but it was a bigger change so i just kept this one focussed. All good, thanks for merging, it's one less vuln (and dependency) on the project :D

boyney123:
Yeah exactly, every little thing helps 🙏

PR #2627, opened 2026-06-18
feat(linter): validate ADRs (Architecture Decision Records)

Vinceveve:
## Motivation

The linter advertises *“comprehensive coverage … supports all EventCatalog resource types”*, but ADRs were silently ignored — there was no schema, no scanner glob, and no entry in the `schemas` registry for them.

The practical consequence: invalid ADR frontmatter passes the linter but breaks the build. For example an ADR with `status: draft`:

```
✔ No problems found!   # npx @eventcatalog/linter
```

```
[InvalidContentEntryDataError] adrs → … data does not match collection schema.
  status: Invalid option: expected one of "proposed"|"accepted"|"rejected"|"deprecated"|"superseded"
# eventcatalog build  → exit 1
```

So the lint gate is green and the break only surfaces post-merge, during `eventcatalog build`. This makes the linter unsuitable as a pre-merge guard for any catalog that uses ADRs.

## What this does

Adds `adr` as a first-class resource type, mirroring the `adrs` content collection in `packages/core/eventcatalog/src/content.config.ts`:

- **Schema** (`src/schemas/adr.ts`): `status` enum (`proposed`/`accepted`/`rejected`/`deprecated`/`superseded`), `date`, optional `decisionMakers`, typed `appliesTo` pointers, and `supersedes`/`supersededBy`/`amends`/`amendedBy`/`related`. The status enum values are kept in sync with core's `ADR_STATUS_VALUES`.
- **Scanner**: discovers `**/adrs/*/index.{md,mdx}` and the versioned variant.
- **Registry**: registered in `schemas` so it's picked up by schema validation; owners references are validated by the existing generic owners handling.
- **Best practices**: ADRs are exempt from `owner-required` because they declare ownership via `decisionMakers` (`owners` is optional in the schema) — otherwise the linter would reject valid ADRs the build accepts.

After this change, the `status: draft` example above is reported as a `schema/required-fields` error (exit 1), matching the build.

## Tests

- `schema-validator.test.ts`: valid ADR, every valid status value, the invalid-status regression, missing required fields, typed `appliesTo`/`decisionMakers`/relationships, and an unknown `appliesTo` type.
- `scanner.test.ts`: ADR id extraction at domain/service level and versioned.
- `new-rules.test.ts`: ADRs without `owners` are not flagged for `owner-required` (and a service still is).

All 155 linter tests pass.

> Note: this touches only `@eventcatalog/linter`; no change is needed in `eventcatalog/eventcatalog-editor`.

### DIFF
--- a/.changeset/linter-validate-adrs.md
+++ b/.changeset/linter-validate-adrs.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/linter": minor
+---
+
+Validate ADRs (Architecture Decision Records). The linter now scans `**/adrs/*/index.{md,mdx}` and validates their frontmatter against a schema mirroring the EventCatalog content collection — including the `status` enum (`proposed`, `accepted`, `rejected`, `deprecated`, `superseded`). Previously invalid ADR frontmatter (e.g. an unknown `status`) passed the linter but failed `eventcatalog build` with `InvalidContentEntryDataError`. ADRs are exempt from the `owner-required` best practice since they declare ownership via `decisionMakers`.
--- a/packages/linter/README.md
+++ b/packages/linter/README.md
@@ -31,6 +31,7 @@ A comprehensive linter for EventCatalog that validates frontmatter schemas and r
 - 📊 **Entities**
 - 👤 **Users**
 - 👥 **Teams**
+- 📝 **ADRs** (Architecture Decision Records)
 
 ## 📦 Installation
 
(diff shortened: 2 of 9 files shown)


boyney123:
Thank you @Vinceveve ! I had this actually on my list of things to update/push yesterday (funny enough), the branch was not pushed, but I just did https://github.com/event-catalog/eventcatalog/pull/2628

which covers ADR documents, thanks for your PR will merge these into the linter to help

Issue #2635, opened 2026-06-19
Large domain detail page renders the whole descendant graph inline (client:only) — ~27s SSR render + ~8.3MB HTML, can take the server down


Vinceveve:
### Tried this

Running EventCatalog in SSR mode (`output: 'server'`, `@astrojs/node`) and opening the **detail page of a large domain** — a domain that contains sub-domains with many services/messages (in our catalog: 11 sub-domains, ~337 services, ~3368 queries, ~64 events ≈ 3800 graph nodes).

Measured the page server-side (TTFB via `curl`, render done by the node server) and compared with a normal domain.

### This happened

The large domain detail page (`/docs/domains/<id>/<version>`) is pathologically heavy:

| Domain | SSR render (TTFB) | HTML size |
|---|---|---|
| normal domain | ~0.3 s | ~330 KB |
| large domain | **~27 s** | **~8.3 MB** (99% is a single inline line) |

Root cause (in `@eventcatalog/core` 3.45.0):

- `pages/docs/[type]/[id]/[version]/index.astro` **auto-pushes the page NodeGraph** (~L364) with **no opt-out**: it does not check `data.visualiser`, and the `<NodeGraph>` render (~L571) is **not** gated by `isVisualiserEnabled()` (only the "Open in Visualiser" link is, ~L585).
- `utils/node-graphs/domains-node-graph.ts` expands the domain **and all sub-domains → all their services → all their messages**, with no node cap / collapse.
- `components/MDX/NodeGraph/NodeGraph.astro` renders `<AstroNodeGraph client:only="react" nodes={nodesWithLayout} edges={edges} />` — **the layout is computed server-side** (the ~27 s) and **the full nodes/edges set is serialized inline** into the HTML (the ~8.3 MB single line).

Because in SSR the ~27 s render is **synchronous on the single-threaded Node event loop**, the server cannot answer **anything else during that window — including health probes**. In Kubernetes this makes the liveness probe fail → the pod is killed → other users get `503`. So one view of a large-domain page can take the whole instance down.

### Expected this

Opening a large domain page should not produce a multi-second blocking render nor a multi-MB inline payload, and should not be able to take the server down. Expected one or more of:

- `visualiser: false` (frontmatter) and/or `visualiser.enabled: false` (config) **honored on the detail-page graph**, not only on `/visualiser/*` pages and the sidebar.
- A **node cap** with a "graph too large — click to load" fallback for large domains.
[...]

PR #2639, opened 2026-06-23
Defer dagre layout for composed sub-graphs and use map-based de-duplication

boyney123:
## What This PR Does

Improves node graph generation performance. The domains node graph composes other graphs (services, agents, data products, sub-domains) by reusing their `getNodesAndEdges` helpers. Previously each sub-graph ran the expensive `dagre.layout()` pass independently, and the parent ran it again — wasting work. This PR adds a `layout` option so sub-graphs skip layout and the parent runs a single layout pass at the end. It also replaces O(n²) `find`-based de-duplication of nodes and edges with O(n) `Map`-based lookups.

### Key Changes
- Add `layout?: boolean` (default `true`) to the `getNodesAndEdges` props for `services`, `domains`, and `data-products` node graphs.
- `domains-node-graph` now passes `layout: false` to each composed sub-graph and runs a single `dagre.layout(flow)` after all nodes/edges are collected.
- Replace `Array.reduce` + `find` de-duplication of edges (and nodes, in services) with `Map`-based de-duplication.

## How It Works

`getNodesAndEdges` gates the `dagre.layout(flow)` call behind the new `layout` flag. When the domains graph builds its composite layout, it calls the service/agent/data-product/sub-domain helpers with `layout: false` so they only contribute nodes and edges to the shared `flow` without laying it out. After everything is collected, the domains graph performs one layout pass. Behavior for standalone graphs is unchanged because `layout` defaults to `true`.

The de-duplication change keys nodes and edges by `id` in a `Map`, avoiding the repeated linear `find` scans. The services graph preserves its existing custom-color merge logic when duplicate edges converge.

## Breaking Changes

None. The new option is additive and defaults preserve existing behavior.

## Additional Notes

No editor (`eventcatalog/eventcatalog-editor`) change is needed — these utilities are internal to `@eventcatalog/core`.

### DIFF
--- a/.changeset/quiet-otters-jump.md
+++ b/.changeset/quiet-otters-jump.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+Improve node graph performance by deferring dagre layout when composing sub-graphs and using Map-based de-duplication for nodes and edges
--- a/packages/core/eventcatalog/src/utils/node-graphs/data-products-node-graph.ts
+++ b/packages/core/eventcatalog/src/utils/node-graphs/data-products-node-graph.ts
@@ -26,6 +26,7 @@ interface Props {
   version: string;
   defaultFlow?: DagreGraph;
   mode?: 'simple' | 'full';
+  layout?: boolean;
 }
 
 const getNodePropertyFromCollectionType = (type: string) => {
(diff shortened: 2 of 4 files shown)


boyney123:
Related https://github.com/event-catalog/eventcatalog/issues/2635

Issue #2643, opened 2026-06-29
EventCatalog v4 - Roadmap


boyney123:
Hi folks,

This just a public list of things I'm working on preparing for V4 of EventCatalog. You will soon be able to get access to the beta to upgrade and test new features.

You can try it out today

```
npm i @eventcatalog/core@beta
```

### Why v4?
There are a few changes I've been wanting to make for awhile, adding some new resources, simplifying EventCatalog business model and license. Nothing is necessary breaking changes, but a fair amount of internal moving parts, so I think a major release here makes sense.

### Changes coming to v4

- [x] Adding support for [Systems](https://github.com/event-catalog/eventcatalog/issues/2539) and System Context Maps. This new resource will let you group entities together in ways that were not possible before in EventCatalog. At the moment you have to group resources in Domains, but some things require a different level of grouping (see that issue). With systems you can define a collection of resources that make a system, for example a legacy system which may consist of many services, databases, and messages. The Context diagram will give you a similar experience like the C4 Model. Allowing you to view your domains/systems at a high level and drill down into details.

- [ ] Allow to link commands and events. This will give people the ability to link messages directly together helping you understand which command triggers which event. More details can be found in the issue https://github.com/event-catalog/eventcatalog/issues/2263

- [ ] Work toward new and simpler patterns to Sync schemas into EventCatalog. Users will be able to reference a schema in the front matter of the message itself. EventCatalog will fetch your schema at buildtime or runtime and display this schema for you. For example you can say something like `schema: apicurio://my-group/schema.json` which will fetch it from apicurio (or any target).

- [x] Introducing [EventCatalog Agents](https://github.com/event-catalog/agents). EventCatalog is a model of your architecture defined through markdown and resources. We can use this to help with automation and guardrails in your systems. For example detecting breaking changes, notifying consumers, and also transforming your code into docs. We already started making progress here, but this will be included in v4 release https://github.com/event-catalog/agents

- [x] Merging our "explore/discover" (table pages) together and allow filtering on "kind" similar to how backstage works with different resources. Our Vertical navigation bar is getting busy with the amount of resources. We can simplify this UI by merging or having a main page to then dive into the sub pages cleaning up the user experience.

- [x] Updating EventCatalog to Astro 7 (rust runtime/build time features) increasing builds times and performance on EventCatalog.

[...]

PR #2646, opened 2026-06-29
Add Kind filter to Discover and simplify sidebar navigation

boyney123:
## What This PR Does

Replaces the many resource-specific "Browse" links in the sidebar with a single **Discover** entry, and adds a **Kind** filter to the discover pages so users can switch between the available resource collections (domains, systems, services, events, commands, queries, flows, etc.) without leaving the page. Collections with zero resources are hidden from both the sidebar target and the Kind filter. Also includes a `CopyAsMarkdown` component, Prompt component updates, and a large batch of default example-catalog content (entities, docs).

### Key Changes

- **Discover Kind filter**: New `CollectionKindOption` type and `collectionKinds` prop on `DiscoverTable`. When more than one collection has resources, a "Kind" dropdown renders at the top of the filters and navigates between `/discover/<kind>` pages.
- **FilterDropdown**: Added a `clearable` prop so the Kind dropdown can hide the clear affordance for its always-selected current value.
- **Sidebar navigation**: `VerticalSideBarLayout` now collapses the per-resource Browse links into a single Discover link that points at the first non-empty collection. "Catalog" renamed to "Home".
- **Icon map**: Added several Lucide icons (BotMessageSquare, ClipboardCheck, FileWarning, ListChecks, Radar, Radio, Route, ShieldCheck, Sparkles).
- **Supporting changes**: `CopyAsMarkdown.tsx` component, Prompt MDX component updates, custom-documentation tweaks, and docs page adjustments.
- **Example catalog**: Adds entities and docs across the default example domains as part of the ongoing default-catalog rebuild.

## How It Works

The discover page (`/discover/[type]`) builds a `collectionKinds` array of all resource types with a count and a `buildUrl` href, filtered to collections that actually have resources. This array is passed to `DiscoverTable`, which renders a Kind `FilterDropdown` when more than one kind is present. Selecting a kind sets `window.location.href` to that collection's discover page. The sidebar derives the same set of non-empty collections to pick a single default Discover target.

## Breaking Changes

None.

## Additional Notes

- Changeset included: `.changeset/discover-kind-navigation.md` (patch).
- No editor-repo (`eventcatalog/eventcatalog-editor`) change is needed for this PR.


boyney123:
https://github.com/event-catalog/eventcatalog/issues/2643

PR #2649, opened 2026-06-30
perf(schema-explorer): stop inlining the full payload into client:load props

### LINKED ISSUE #2648: Schema Explorer serializes the full schemas array as client:load props → very large HTML on big catalogs (3.48.4 pagination is render-only)
## Problem

`/schemas/explorer` renders `<SchemaExplorer client:load schemas={schemas} />` (`pages/schemas/explorer/index.astro`). `schemas` (built in `schemas/explorer/_index.data.ts`) is the **complete** list of every message/spec/contract, and each item carries its full `schemaContent` **plus** `data.producers`, `data.consumers` and `examples`.

Because the island is `client:load`, Astro serializes the **entire array** into the HTML for hydration. On a large catalog (tens of thousands of resources) this produces a very large HTML document — on our catalog ~**80MB** (~4MB gzipped). The server render + gzip are fast, but the browser has to inflate + hydrate the whole payload, so the tab hangs / times out.

**3.48.4** added client-side pagination (`ITEMS_PER_PAGE`, `paginatedMessages`), which limits what is *rendered*, but the full `schemas` array is still serialized as props — so the payload size is unchanged.

## Where the weight is

Measured on our catalog, the dominant cost was **not** the schema bodies (~10% of the payload) but the per-item **`producers` / `consumers` / `examples`** arrays (`{id, version}` tuples) inlined for every item across a densely cross-referenced catalog.

## Proposed fix

The list view only needs `id` / `name` / `version` / `summary` / `collection` / `schemaExtension` to filter, paginate and select. So:

1. In `_index.data.ts`, **don't inline `producers`/`consumers`/`examples`** into the list props. The relationships already live on each resource's own page, and the detail-panel sections are conditional on `length > 0`, so they degrade cleanly to hidden.
2. **Lazy-fetch the selected message's `schemaContent`** from the existing `/api/schemas/<collection>/<id>/<version>` route when an item is opened, instead of inlining every body.

This keeps the page small regardless of catalog size and composes with the new pagination.

We run exactly this as a downstream `patch-package` patch on a ~70k-resource catalog: `/schemas/explorer` went from ~**80MB → ~1.2MB** and renders instantly. Happy to open a PR (incoming).

Related to #2635 — same "serialize-everything-inline" class of problem surfacing on large catalogs.

Vinceveve:
Closes #2648 (related to #2635).

## Problem

`/schemas/explorer` renders `<SchemaExplorer client:load schemas={schemas} />`, and `schemas` carries, for **every** item, its full `schemaContent` plus `data.producers`, `data.consumers` and `examples`. Because the island is `client:load`, Astro serializes the **entire** array into the HTML. On large catalogs this becomes a multi-MB document that the browser can't inflate/hydrate without hanging. The 3.48.x client-side pagination limits what's *rendered* but not what's *serialized*, so the payload is unchanged.

## Change

- `schemas/explorer/_index.data.ts`: stop inlining `producers`/`consumers`/`examples` into the list props (they're only shown for the *selected* item, the panel sections are conditional on `length > 0`, and the full relationships live on each resource's own page), and stop inlining message `schemaContent`.
- `SchemaExplorer.tsx`: lazy-fetch the **selected** message's content from the existing `/api/schemas/<collection>/<id>/<version>` route, and pass it to `SchemaDetailsPanel`. Specs/contracts that still carry inline content are used as-is (early return).

The list view only needs `id`/`name`/`version`/`summary`/`collection`/`schemaExtension` to filter, paginate and select — so the heavy per-item data no longer needs to ship up-front. Composes with the existing pagination.

## Result

Validated downstream (via `patch-package`) on a large (~70k-resource) catalog: `/schemas/explorer` dropped from ~80MB to ~1.2MB of HTML and renders instantly; selecting a message fetches its body on demand.

## Notes

- A changeset is included (`patch`).
- I haven't been able to run the full monorepo build here — please let CI validate. The change is behaviour-preserving for the selected-item view (content arrives via the API route) and only removes data that wasn't displayed in the list.

### DIFF
--- a/.changeset/schema-explorer-inline-payload.md
+++ b/.changeset/schema-explorer-inline-payload.md
@@ -2,4 +2,4 @@
 '@eventcatalog/core': patch
 ---
 
-Schema Explorer: stop inlining the full payload into the `client:load` props. `producers`/`consumers`/`examples` are no longer inlined per item, and a selected message's schema body is lazy-fetched from `/api/schemas/<collection>/<id>/<version>` instead of being serialized for every schema. Keeps `/schemas/explorer` small (and the browser responsive) on large catalogs, complementing the existing pagination.
+Schema Explorer: stop inlining the full payload into the `client:load` props. `producers`/`consumers`/`examples` are no longer inlined per item, and a selected message's schema body is lazy-fetched from an internal explorer route (`/schemas/explorer/content/<collection>/<id>/<version>`) instead of being serialized for every schema. The route ships with every catalog (prerendered in static output, on-demand in SSR), so this is not coupled to the Scale `/api/schemas` API. Keeps `/schemas/explorer` small (and the browser responsive) on large catalogs, complementing the existing pagination.
--- a/.changeset/schema-explorer-inline-payload.md
+++ b/.changeset/schema-explorer-inline-payload.md
@@ -1,0 +1,5 @@
+---
+'@eventcatalog/core': patch
+---
+
+Schema Explorer: stop inlining the full payload into the `client:load` props. `producers`/`consumers`/`examples` are no longer inlined per item, and a selected message's schema body is lazy-fetched from `/api/schemas/<collection>/<id>/<version>` instead of being serialized for every schema. Keeps `/schemas/explorer` small (and the browser responsive) on large catalogs, complementing the existing pagination.
--- a/packages/core/eventcatalog/src/__tests__/api/schema-explorer-content-route.spec.ts
+++ b/packages/core/eventcatalog/src/__tests__/api/schema-explorer-content-route.spec.ts
@@ -1,0 +1,91 @@
+import type { CollectionKey } from 'astro:content';
+import { expect, describe, it, vi, beforeEach } from 'vitest';
+import { getStaticPaths, GET } from '../../pages/schemas/explorer/content/[collection]/[id]/[version].ts';
+import path from 'path';
+import fs from 'node:fs';
+
+const __dirname = path.dirname(new URL(import.meta.url).pathname);
+
+let mockSchemas: any[] = [];
+
+const schemaContent = (fileName: string) => fs.readFileSync(path.join(__dirname, 'schemas', fileName), 'utf8');
+
+const schemaEntry = ({
+  collection,
+  id,
+  version,
+  file,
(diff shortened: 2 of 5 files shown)


boyney123:
Thanks for digging into this and for the detailed write-up. I agree with the root cause here: pagination helps render cost, but the `client:load` island still serializes the full `schemas` array into the HTML, so large catalogs still pay the hydration payload cost.

One thing we need to think through before merging is the lazy-load route. The PR currently fetches selected message content from `/api/schemas/<collection>/<id>/<version>`, but that Schema API is a documented Scale feature and the routes are only injected when Scale is enabled. In static/non-Scale catalogs this would remove inline `schemaContent` from the explorer props, then the lazy fetch would 404/not exist, so the explorer would lose schema content for those users.

I also think the fetch should go through `buildUrl(...)` so catalogs deployed under a base path continue to work.

My current leaning is that we should keep `/api/schemas/...` as the paid public API, and add a separate internal/undocumented explorer data route for this page's own hydration, for example something under `/schemas/explorer/_content/...` or `/_eventcatalog/schema-explorer/...`. That route would be treated as implementation detail, work in static and server output, and only return the data the explorer already exposes today, but on demand rather than all upfront.

Going to think about the exact shape a bit more and come back with a concrete path. The performance direction here still looks right; we just need to avoid coupling the free explorer to the Scale Schema API.

Vinceveve:
Great points, thanks @boyney123 

> - **No more Scale coupling.** Confirmed `/api/schemas/...` is only injected under `isEventCatalogScaleEnabled()`, so I moved the lazy fetch to a new internal explorer route: `/schemas/explorer/content/<collection>/<id>/<version>`. It ships with every catalog (no plan gating), prerenders to a static file per message in static output, and resolves on demand in SSR — following the same pattern as the existing `.mermaid.ts` endpoint.
> - **Base path.** The fetch now goes through `buildUrl(...)`.

Is it better like this, do I miss something ?

boyney123:
thanks for this @Vinceveve sorry it took so long, I merged your changes in with https://github.com/event-catalog/eventcatalog/pull/2858, as I was doing some other stuff there anyway. Hope thats OK.

LMK if it made any improvemnets for you

Issue #2669, opened 2026-07-06
Support custom documentation in the EventCatalog MCP server


boyney123:
## Motivation

EventCatalog supports bring your own documentation through custom docs/pages under `docs/`, but the MCP server currently focuses on catalog resources such as domains, services, messages, flows, teams, and users.

This means MCP clients like Claude Desktop, Cursor, Windsurf, and other AI tools can query the architecture model, but they cannot reliably discover or retrieve the supporting custom documentation that teams often use for runbooks, onboarding guides, standards, operating models, and internal architecture guidance.

For many catalogs, this custom documentation is essential context for answering architecture questions correctly.

## Proposal

Add first-class MCP support for custom documentation.

The MCP server should expose custom docs through tools and/or resources so MCP clients can:

- List available custom documentation pages.
- Search custom documentation by title, summary, path, and markdown content.
- Retrieve a specific custom documentation page.
- Include useful metadata such as title, summary, owners, badges, path, slug, and markdown content.
- Support nested custom docs under `docs/**`.
- Respect existing MCP availability, auth, and Scale-plan behavior.

## Suggested MCP surface

Potential tools:

- `getCustomDocs`
  - Lists custom documentation pages.
  - Supports pagination and optional search.
  - Optionally supports filtering by path/directory.

- `getCustomDoc`
  - Retrieves a custom documentation page by path, slug, or id.
  - Returns frontmatter metadata and markdown content.

Potential resources:

- `eventcatalog://docs`
- `eventcatalog://docs/{path}`

## Acceptance criteria

- Custom docs from `docs/**/*.{md,mdx}` are discoverable through the MCP server.
- Individual custom docs can be retrieved with their metadata and markdown body.
- Search works across at least title, summary, path, and markdown body.
- Nested documentation paths are supported.
- MCP health/resource metadata includes custom documentation where appropriate.
- Existing MCP tools and resources continue to work unchanged.
- Tests cover listing, searching, retrieving, missing documents, and nested custom docs.
- Documentation is updated to explain how custom docs are exposed to MCP clients.

## Notes

Relevant existing areas:

- MCP server: `packages/core/eventcatalog/src/enterprise/mcp/mcp-server.ts`
- MCP tests: `packages/core/eventcatalog/src/enterprise/mcp/__tests__/mcp-server.spec.ts`
- Custom docs SDK helpers: `packages/sdk/src/custom-docs.ts`
[...]

PR #2670, opened 2026-07-06
Add custom documentation tools to AI Chat and MCP Server

boyney123:
## What This PR Does

Exposes EventCatalog's custom documentation pages (guides, runbooks, architecture notes, onboarding docs) to both the AI Chat and the MCP Server through three new tools. This lets AI assistants discover, search, and read the free-form documentation that lives in the catalog but isn't captured by the structured resource tools. Full-text search is powered by MiniSearch and operates at the markdown-section level, so results point at the specific part of a page rather than the whole document.

### Key Changes

- New shared tool implementations in `catalog-tools.ts`:
  - `getCustomDocs` — lists custom documentation pages (metadata only), with cursor pagination and an optional case-insensitive filter over title/id/summary.
  - `searchCustomDocs` — section-level full-text (BM25) search returning the best matching sections with page id, heading, snippet, and score.
  - `getCustomDoc` — returns the full markdown of a page by id or slug, or just a single section when a heading is provided.
- New `docs-search.ts` utility with:
  - `chunkMarkdownByHeadings` — splits markdown into section chunks at ATX headings (ignoring headings inside fenced code blocks).
  - `extractSnippet` — builds a query-centred excerpt for search results.
  - `buildDocsSearchIndex` — constructs an in-memory MiniSearch index, one entry per section, with field boosts and prefix matching.
- Registered all three tools in both the AI Chat API (`chat-api.ts`) and the MCP Server (`mcp-server.ts`) with matching Zod input schemas and tool descriptions.
- Added `minisearch` (`^7.2.0`) as a dependency of `@eventcatalog/core`.
- Added unit tests for the search utilities and the catalog tools, plus MCP server registration coverage.

## How It Works

Custom documentation pages come from the `customPages` collection. The list and get tools read directly from the collection; the search tool lazily builds a MiniSearch index (cached unless `DISABLE_EVENTCATALOG_CACHE=true`) that indexes each page as one entry per markdown section. Search matches are boosted by title/heading/summary and support prefix matching, and each hit is returned with a snippet centred on the query. `getCustomDoc` can return either the whole page or a single section, keeping responses small for large pages.

## Breaking Changes

None.

## Additional Notes

- Tools live in `@enterprise/tools/catalog-tools.ts` following the shared-implementation pattern, so AI Chat and MCP Server stay in sync.
- No changes are needed in the editor repo (`eventcatalog/eventcatalog-editor`) — this is server-side tooling for AI Chat and the MCP Server only.

### DIFF
--- a/.changeset/curious-otters-search.md
+++ b/.changeset/curious-otters-search.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": patch
+---
+
+Add custom documentation tools (getCustomDocs, searchCustomDocs, getCustomDoc) to AI Chat and MCP Server, with section-level full-text search powered by MiniSearch
--- a/packages/core/eventcatalog/src/enterprise/ai/chat-api.ts
+++ b/packages/core/eventcatalog/src/enterprise/ai/chat-api.ts
@@ -18,6 +18,9 @@ import {
   getDataProductInputs as getDataProductInputsImpl,
   getDataProductOutputs as getDataProductOutputsImpl,
   getArchitectureDiagramAsMermaid as getArchitectureDiagramImpl,
+  getCustomDocs as getCustomDocsImpl,
+  searchCustomDocs as searchCustomDocsImpl,
+  getCustomDoc as getCustomDocImpl,
   collectionSchema,
   resourceCollectionSchema,
   messageCollectionSchema,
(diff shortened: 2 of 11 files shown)


boyney123:
https://github.com/event-catalog/eventcatalog/issues/2669

Issue #2747, opened 2026-08-03
Built SSR server crashes at runtime: "The requested module 'neotraverse' does not provide an export named 'forEach'"


freaz:
## Describe the bug

After upgrading to `@eventcatalog/core@4.3.4`, the production SSR server (`eventcatalog build` + `eventcatalog start`) returns HTTP 500 on every page with:

```
[ERROR] file:///app/dist/server/chunks/_astro_content_DnxXjE09.mjs:9
import { forEach } from "neotraverse";
         ^^^^^^^
SyntaxError: The requested module 'neotraverse' does not provide an export named 'forEach'
    at async ModuleJob.run (node:internal/modules/esm/module_job:431:5)
    at async AppPipeline.getModuleForRoute (file:///app/dist/server/entry.mjs:5326:11)
    ...
```

## Root cause

`@eventcatalog/core@4.3.4` ends up with **two astro majors** in the dependency tree:

- `@eventcatalog/core` itself depends on `astro@^7.1.1` (resolved 7.1.6), which depends on `neotraverse@^1.0.1` and emits `import { forEach } from "neotraverse"` in the built server chunks — the named `forEach` root export only exists in neotraverse **1.x**.
- `astro-compress@^2.4.0` (`astro: *`) and `auth-astro@^4.2.0` (`astro: >=4.0.0`) resolve a **second** `astro@6.4.8`, which depends on `neotraverse@^0.6.18`.

npm hoists astro 6.4.8's `neotraverse@0.6.18` to the top-level `node_modules`, and nests `neotraverse@1.0.1` under `@eventcatalog/core/node_modules`:

```
└─┬ @eventcatalog/core@4.3.4
  ├─┬ astro-compress@2.4.1
  │ └─┬ astro@6.4.8
  │   └── neotraverse@0.6.18   <-- hoisted to top-level node_modules
  ├─┬ astro@7.1.6
  │ └── neotraverse@1.0.1      <-- nested, not visible to dist/
  └─┬ auth-astro@4.2.0
    └── astro@6.4.8 deduped
```

The built output in `dist/server` keeps `neotraverse` as an external bare import. At runtime Node resolves it from the project root's `node_modules` — landing on the hoisted **0.6.18**, whose root export is the legacy default-only API (no named `forEach`), so the server crashes on the first request.

Whether this bites depends on npm's hoisting outcome, so it can appear/disappear across lockfile updates — which makes it a nasty one to track down.

## Steps to reproduce

[...]

PR #2748, opened 2026-08-04
Make astro-compress an optional dependency

boyney123:
## What This PR Does

Removes `astro-compress` from the `@eventcatalog/core` dependency list and loads it lazily from the user's own EventCatalog project instead. Compression is off by default, so every catalog was paying the install cost of a package that most builds never used. Catalogs that want `compress: true` now install `astro-compress` themselves.

Also fixes the sidebar so top-level groups always stay expanded regardless of how many children they have.

### Key Changes

- Dropped `astro-compress` from `packages/core/package.json` dependencies (188-line reduction in `pnpm-lock.yaml`).
- Added `packages/core/eventcatalog/src/plugins/astro-compress.ts`, which resolves `astro-compress` from the user's project directory and throws an actionable install message when it's missing.
- `astro.config.mjs` now calls `loadAstroCompressIntegration(projectDirectory)` instead of a bare `import('astro-compress')`.
- `canCollapseGroup()` takes an `isTopLevel` flag and returns `false` for top-level sidebar groups.
- Documented the new install step for compression, tagged `<AddedIn version="4.4.0" />`.

## How It Works

The integration is resolved with `createRequire(join(projectDirectory, 'package.json'))` so Node looks for `astro-compress` in the *catalog's* `node_modules`, not EventCatalog's. The resolved path is converted to a file URL before the dynamic `import()` so it works on Windows, and the `/* @vite-ignore */` hint stops Vite from trying to statically analyse the specifier.

Two failure modes are handled explicitly: the package isn't installed (resolution throws, rethrown with install instructions), and the package is installed but doesn't export a default integration function.

Compression options are unchanged — `Logger: 0` and `CSS: false` — so catalogs that add the dependency get identical output to before.

## Breaking Changes

**Yes, for catalogs using `compress: true` on static builds.** These will now fail the build with an install message until `astro-compress` is added as a dev dependency:

```bash
npm install --save-dev astro-compress
```

Catalogs with compression disabled (the default since v2.61.9) and all SSR builds are unaffected. Shipping as `minor` per the changeset.

## Additional Notes

- New unit tests in `astro-compress.test.ts` cover both the happy path and the missing-package error by scaffolding a throwaway project dir with a stub package. Full suite for the touched files: 11 passing.
- The docs pre-commit hook syncs `packages/core/docs` from the sibling `website-2` repo. That sync reverts three unrelated files (`98-versioning-resources.md`, `01-version-services.md`, `03-features.md`) because the versionless-URL docs added in 04e5d190 were written straight into `packages/core/docs` and never landed in website-2. This is pre-existing drift, not something this PR introduces — but it means those `4.3.3` docs need re-adding to website-2 to stop the hook undoing them.
- The compression doc edits were made in `website-2` (the source of truth) and are **uncommitted in that sibling repo** — they need committing and PRing there separately.
- `pnpm run check` currently fails on a stale build artifact in the git-ignored `examples/default/dist/` directory (`FraudCheckCompleted` has a non-`x-` prefixed `tags` property). Unrelated to this change.

### DIFF
--- a/.changeset/quiet-otters-tap.md
+++ b/.changeset/quiet-otters-tap.md
@@ -1,0 +1,5 @@
+---
+"@eventcatalog/core": minor
+---
+
+feat(core): astro-compress is now an optional peer dependency, install it in your catalog to use `compress: true`. Top level sidebar groups now stay expanded.
--- a/packages/core/docs/api/02-config.md
+++ b/packages/core/docs/api/02-config.md
@@ -704,11 +704,19 @@ module.exports = {
 
 - Type: `boolean`
 
-Setting this to true will automatically compress all your CSS, HTML, SVG, JavaScript, JSON and image files in the Astro outDir folder.
+Setting this to true will automatically compress your HTML, SVG, JavaScript, JSON and image files in the Astro outDir folder.
 
 This is disabled by default from EventCatalog v2.61.9.
 
 **This only works for static builds.**
+
+<AddedIn version="4.4.0" />
+
+`astro-compress` is no longer bundled with EventCatalog. Install it in your EventCatalog project before enabling compression:
+
+```bash
+npm install --save-dev astro-compress
+```
 
 ```js title="eventcatalog.config.js"
 module.exports = {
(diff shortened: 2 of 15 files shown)


boyney123:
I believe this should also fix https://github.com/event-catalog/eventcatalog/issues/2747

Issue #943, opened 2024-11-11
Create visualizer pages for channels


boyney123:
### Use Case

The visualizer supports channels (soon) between messages and services, it would be nice to have a visualizer page for channels themself.

Here is a random proposal, just for reference when the feature gets picked up

<img width="859" alt="Pasted Graphic 8" src="https://github.com/user-attachments/assets/85f2ab27-2c55-436a-80d3-dbe48c481751">

### Community Notes

* Please vote by adding a 👍 reaction to the issue to help us prioritize.
* If you are interested to work on this issue, please leave a comment.
* If this issue is labeled **needs-discussion**, it means the spec has not been finalized yet. Please reach out in the [EventCatalog Discord](https://discord.com/invite/3rjaZMmrAm).

boyney123:
Going to close this for now, not planned.

PR #2756, opened 2026-08-09
feat(core): render node graphs for channels

jonaslagoni:
## Motivation

Clicking **Focus node** on a channel node in the visualiser takes you to a 404.

Channel nodes are rendered in message, service and domain graphs, and `buildContextMenuForResource` gives them a **Focus node** link to `/visualiser/channels/{id}/{version}`. That route was never generated: `channels` is missing from the `itemTypes` list in the visualiser `getStaticPaths`, so the page does not exist and the link dead-ends.

The same gap meant `<NodeGraph />` on a channel documentation page rendered empty, since `NodeGraph.astro` only builds a graph when the collection is present in its `getNodesAndEdgesFunctions` map, and `channels` was not.

This PR registers `channels` in both places and adds the missing `getNodesAndEdges` implementation for channels, so the link resolves to a real graph.

The graph shows:

- the producers and the messages that publish into the channel
- the services and agents that consume from it
- any channels it routes to, or is routed from

Messages are picked up whether the channel is declared on the service `sends`/`receives` pointer or on the message itself.

Related to #943, which asked for visualiser pages for channels and was closed without a linked PR.


jonaslagoni:
@boyney123 gentle poke.

Would be really useful to get this fixed as right now we cant easily inspect who consumes and produces to our channel. Been using this quite a few times.

boyney123:
Thanks @jonaslagoni will take a look today hopefully (was away for a few days), thanks for the PR and adding this

boyney123:
Thanks again @jonaslagoni , I added a few more things to this PR

- Looks our example catalog did not have any channels at all, so I added some
- There is some traversal logic from channel to channel stuff that was missing so I added that too in the node graph logic
- Sidebar for channels was missing the "Map" link (as we didnt support it, but this is now added)
- Few bugs with channel nodes that were also fixed (right clicking on them and correct link urls etc).

Think this is in a good place to ship and improve as we find more things in the future.

Thanks for your time on it ❤️

